PR #48, opened 2024-12-24
Fix windows ESM issue

### LINKED ISSUE #47: Error code: ‘ERR_UNSUPPORTED_ESM_URL_SCHEME’ when trying to generate a report
https://github.com/orgs/allure-framework/discussions/2831

epszaw:
fixes #47


packages/core/src/utils/path.ts
```diff
@@ -0,0 +1,20 @@
+import { platform } from "node:os";
+import { win32 } from "node:path";
+
+/**
+ * The function appends `file:///` protocol to the absolute paths on Windows due to ES modules imports specifics
+ * @param path
+ */
+export const normalizeImportPath = (path: string) => {
```
delatrie, line 7: Better to normalize paths with `url.pathToFileURL` as [Node.js docs suggests](https://nodejs.org/api/esm.html#file-urls). We should use it even on mac/linux because node requires module specifiers to be percent-encoded, which `pathToFileURL` does.

delatrie (changes requested):

packages/core/src/utils/path.ts
epszaw: 👍

packages/core/src/utils/module.ts
```diff
@@ -1,9 +1,12 @@
...
 
+const require = createRequire(import.meta.url);
+
 /**
  * Dead simple wrapper around import function to make it possible to mock it in the tests
  * @param path
  */
 export const importWrapper = async (path: string) => {
-  return import(normalizeImportPath(path));
+  return import(normalizeImportPath(require.resolve(path)));
```
delatrie, line 11: I would put `pathToFileURL` here. We need it not only on Windows but on Linux/Mac OS as well. Otherwise, it will throw `URI malformed` or `Cannot find module` on paths containing some special characters (e.g., `%`).

```suggestion
  return import(pathToFileURL(require.resolve(path)).href);
```

delatrie (changes requested):

delatrie (approved):
LGTM 👍

Issue #49, opened 2025-01-02
"Copy error message with stacktrace" to clipboard action doesn't work


plagov:
I add the latest available versions of both `allure3` and `plugin-awesome` within my pipeline through the `npm` commands: `npm add allure` and `npm add @allurereport/plugin-awesome`.

When I open the failed test and I want to copy the error message with the stack-trace to the clipboard by clicking the respective "copy" icon (see attached screenshot), then it doesn't copy anything into the clipboard.

![image](https://github.com/user-attachments/assets/9ac71dbf-97f4-41b4-b105-0bd92b62cd95)

todti:
Hi, @plagov! Could you clarify whether the button copies the error message or if nothing at all is added to the clipboard? Logically, the stack trace shouldn't be included when the button is pressed, message only.

Also, could you please let me know your browser/OS and its version? It’s possible that the issue is related to the environment or operating system you’re using.

Thank you!

vplagov:
Hi @todti!
Sorry for not providing all the details about the issue. I'm on Windows 11. My main browser is MS Edge. Nothing gets copied into the clipboard on Edge. I have disabled all ad blockers and retried - same, nothing is copied over.

Also, I tried the same in Firefox. Also clean, no ad blockers, just default browser settings - it's fine, the error message is copied into the clipboard.

---

It might be useful as well.
I'm using Azure DevOps as a build server. I generate a report as a single HTML during my pipeline and publish this report using this Azure Pipeline task: `JakubRumpca.azure-pipelines-html-report.PublishHtmlReport.PublishHtmlReport@1`. So, all the above steps I did with Edge and Firefox are against the report published to the Azure Pipeline build.

vplagov:
And since we are on this "copy to clipboard" topic, this action also doesn't work on Edge and does on Firefox.

![image](https://github.com/user-attachments/assets/523469a2-5011-43f5-99f6-f487ff1a8bd8)

epszaw:
@todti does the fix handle issue with the `fullName` too?

todti:
@epszaw yes, I've fixed it as well

@plagov you can try again now, it should be working. If you have any questions, feel free to reopen this issue.

vplagov:
@todti does the fix work with Edge browser? Doesn't look it is fixed for Edge.
When I click on the icon, I see now the animation and a tooltip text changes to the "Successfully copied", but still nothing in the clipboard.

Same with the full name of the scenario.

[...]

Issue #41, opened 2024-12-20
Make hitory chart clickable


epszaw:
Click on the chart circles should follow to the history entry page

Issue #34, opened 2024-12-20
Add pointer cursor for all interactive elements


epszaw:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #50, opened 2025-01-06
fix: a lot of small fixes and improvements

todti:
Closes:
- #41 
- #39
- #38 
- #35 
- #34 
- #32 
- #30 
- #33 

Fixes:
- fixed Armenian locale name and ISO code
- added fallback for copy to clipboard for old browsers includes Edge #49


packages/web-awesome/src/components/app/TestResult/TestResultHistory/TestResultHistoryItem.tsx
```diff
@@ -9,38 +9,49 @@ import TreeItemIcon from "@/components/app/Tree/TreeItemIcon";
...
+        {Boolean(message) && (
+          <span onClick={() => setIsOpen(!isOpened)}>
+            <ArrowButton isOpened={isOpened} icon={arrowsChevronDown.id} />
+          </span>
+        )}
+        <div
+          className={styles["test-result-history-item-wrap"]}
+          onClick={(e) => {
+            e.stopPropagation();
+            openInNewTab(navigateUrl);
```
epszaw: Let's open the history entry in the same tab when the row has been clicked. In current implementation, it doesn't make sense to have `Open in new tab` button when the row works in the same way

epszaw (changes requested):

packages/web-commons/src/sanitizeHtml.ts
```diff
@@ -0,0 +1,15 @@
...
+const defaultParams: IFilterXSSOptions = {
+  stripIgnoreTagBody: ["script"],
+  whiteList: {
+    div: ["style"],
+    span: ["style"],
+  },
+  css: true,
+};
+
+export const sanitizeHtml = (html: string, params: IFilterXSSOptions = defaultParams) => {
```
epszaw: It's better not to use `IFilterXSSOptions` directly. Keep only a single `html` parameter for now, we add parameters, if they'll be required

epszaw (changes requested):

epszaw (approved):

PR #51, opened 2025-01-07
feat: design system implementation

todti:


packages/web-components/vitest.config.ts
```diff
@@ -0,0 +1,12 @@
+import { createRequire } from "node:module";
```
epszaw, line 1: Actually, we don't have any unit test now. Setup some stub tests for some of components

packages/web-components/vitest.config.ts
```diff
@@ -0,0 +1,12 @@
...
+import { defineConfig } from "vitest/config";
+
+const require = createRequire(import.meta.url);
+
+export default defineConfig({
+  test: {
+    include: ["./test/**/*.test.ts"],
```
epszaw: Change to `./src/**/*.test.ts` because every component will have it's own test file

packages/web-components/package.json
```diff
@@ -0,0 +1,136 @@
...
+    "eslint-plugin-jsdoc": "^50.0.0",
+    "eslint-plugin-jsx-a11y": "^6.10.0",
+    "eslint-plugin-n": "^17.10.1",
+    "eslint-plugin-no-null": "^1.0.2",
+    "eslint-plugin-preact": "^0.1.0",
+    "eslint-plugin-prefer-arrow": "^1.2.3",
+    "eslint-plugin-react": "^7.36.1",
+    "eslint-plugin-react-hooks": "^4.6.2",
+    "eslint-plugin-storybook": "^0.11.1",
+    "filesize": "^10.1.6",
```
epszaw: I'm not sure, that every single dependency is used in the project. Check the dependencies and remove unused ones. You can use `depcheck`, but be careful, it doesn't guarantee, that found packages are really useless

packages/web-components/package.json
```diff
@@ -0,0 +1,136 @@
...
+    "build:ds": "storybook build",
+    "dev:ds": "storybook dev -p 6006",
+    "lint": "eslint --ext ts,tsx --report-unused-disable-directives --max-warnings 0"
+  },
+  "dependencies": {
+    "@allurereport/core-api": "workspace:*",
+    "@allurereport/web-commons": "workspace:*",
+    "@preact/signals": "^1.3.0",
+    "d3-shape": "^3.2.0",
+    "i18next": "^24.0.2",
```
epszaw: If we talk about unnecessary dependencies, `i18next` should be used on the end-project, not on the design system level. Focus on production dependencies

packages/web-components/package.json
```diff
@@ -0,0 +1,136 @@
...
+    "allure",
+    "testing",
+    "design system",
+    "components"
+  ],
+  "repository": "https://github.com/allure-framework/allure3",
+  "license": "Apache-2.0",
+  "author": "Qameta Software",
+  "type": "module",
+  "exports": {
```
epszaw, line 15: Add `fonts` here and check, that the files can be imported in scss and js files

packages/web-components/src/assets/scss/mixins.scss
```diff
@@ -0,0 +1,26 @@
+@mixin status-bg-and-text {
```
epszaw, line 1: Think about the mixins vendoring because we actively use them in the applications

epszaw: For sure, we shouldn't compile them, it should be `scss` files as is. Maybe, we can create set of `css` classes, but it requires more changes than keeping the current approach

packages/web-components/src/assets/scss/code.css
```diff
@@ -0,0 +1,71 @@
+pre[class*="language-"],
```
epszaw: Convert the file to `.scss`

packages/web-components/src/assets/scss/fonts.scss
```diff
@@ -0,0 +1,21 @@
+@font-face {
```
epszaw: Don't include the file to the design system. Connect the fonts in the storybook, to see the proper view of the components and connect the fonts on the end-project level

packages/web-components/src/assets/scss/typography.scss
```diff
@@ -0,0 +1,218 @@
+:root {
+  --font-family: "PTRootUIWebVF", ui-sans-serif, system-ui, sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
```
epszaw, line 2: Add docs to the readme, which fonts we have to install to reach the proper view of the components. And add notes, that we distribute the fonts in the package

packages/web-components/src/assets/svg/arrows-chevron-down.svg
```diff
@@ -0,0 +1,5 @@
+<svg viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
```
epszaw: It would be nice to notice icons in the readme too

epszaw (changes requested):

package.json
```diff
@@ -34,6 +34,7 @@
     "backbone.marionette@npm:3.5.1/underscore": "^1.13.6"
   },
   "devDependencies": {
+    "@preact/preset-vite": "^2.9.4",
```
epszaw: Why do we need it there? Don't we need to move the dependency to the related package?

epszaw (approved):

PR #54, opened 2025-01-09
Improve attachment processing

baev:
1. cache detected content type
2. cache calculated file extensions
3. fix file extension calculation for attachments
4. now all the attachments files in the report have proper file extensions

### DIFF
--- a/packages/reader-api/src/resultFile.ts
+++ b/packages/reader-api/src/resultFile.ts
@@ -1,5 +1,5 @@
 import type { ResultFile } from "@allurereport/plugin-api";
-import { extension, lookup } from "mime-types";
+import { lookup } from "mime-types";
 import {
   ReadStream,
   closeSync,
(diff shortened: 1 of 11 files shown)


packages/reader-api/src/resultFile.ts
```diff
@@ -11,13 +11,16 @@ import {
...
-import { basename, extname } from "node:path";
+import { basename } from "node:path";
 import type { Readable } from "node:stream";
 import { pipeline } from "node:stream/promises";
 import { detectContentType } from "./detect.js";
+import { extension } from "./utils.js";
 
 export abstract class BaseResultFile implements ResultFile {
   fileName: string;
+  extension: string | false = false;
```
epszaw, line 22: Why we don't use `null` or `undefined` there?

epszaw (approved):

packages/reader-api/src/resultFile.ts
baev: I guess `false` is more verbose

PR #55, opened 2025-01-09
fix: allure1 ignores nested steps and attachments of nested steps that consist of a single element

delatrie:
The PR fixes the following issues with the allure1 reader:

  - A single nested step is ignored
  - A single attachment of a nested step is ignored

The reason is the insufficient check performed by the `isArray` callback of `XMLParser`. Since steps may nest arbitrarily, a set of fixed array element paths is insufficient. The PR adds an extra check with regular expressions.

### DIFF
--- a/packages/reader/src/allure1/index.ts
+++ b/packages/reader/src/allure1/index.ts
@@ -51,11 +51,12 @@ const arrayTags: Set<string> = new Set([
   "test-suite.labels.label",
   "test-suite.test-cases.test-case",
   "test-suite.test-cases.test-case.labels.label",
-  "test-suite.test-cases.test-case.attachments.attachment",
   "test-suite.test-cases.test-case.parameters.parameter",
-  "test-suite.test-cases.test-case.steps.step.attachments.attachment",
 ]);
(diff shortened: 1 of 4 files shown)


packages/reader/src/allure1/index.ts
```diff
@@ -51,19 +51,20 @@ const arrayTags: Set<string> = new Set([
...
+];
 
 const xmlParser = new XMLParser({
   parseTagValue: false,
   ignoreAttributes: false,
   attributeNamePrefix: "",
   removeNSPrefix: true,
   allowBooleanAttributes: true,
-  isArray: (tagName, jPath) => arrayTags.has(jPath),
+  isArray: (tagName, jPath) => arrayTags.has(jPath) || arrayTagPatterns.some((p) => p.test(jPath)),
```
baev: Maybe check tag names instead of paths? `attachment`, `step`, `parameter`, 'label', 'test-case' tags are always arrays

delatrie: That makes sense. The format is closed: false positives are unlikely.
And it's much cheaper.

baev (approved):

PR #83, opened 2025-01-31
empty states when no test steps in the testcase

### LINKED ISSUE #81: Add placeholders when there are no steps
It would be better to have a placeholder when test result has an empty body:

Image

todti:
closes #81 
image


packages/web-awesome/src/stores/testResults.ts
```diff
@@ -56,6 +56,7 @@ export const fetchTestResult = async (testResultId: string) => {
       error: undefined,
       loading: false,
     };
+    console.log(testResultStore.value);
```
baev: please remove

baev (changes requested):

baev (approved):

PR #84, opened 2025-01-31
feat: xcresult reader

delatrie:
The PR adds support for xcresult files to Allure 3.

Improvements compared to allure-xcresults:

  - Parametrized tests
  - Destinations (a.k.a., devices)
  - Attachments are chronological
  - Suites (parentSuite = bundle, suite = the top most class, subSuite = nested classes)
  - Packages (the format is `<project>.<bundle>`)
  - Tags
  - Bugs (converted to issue links)
  - expected failures
  - additional API activities (parameters, preconditions, expected results, flaky, muted, known)
  - attachment content type based on UTI
  - attachments from auto-captured screenshots/video

What's left

  - HEIC -> jpeg (eroshenkoam/xcresults#38; check if it's still a problem)
    *Update:* it will be implemented in a separate PR, see #111.

#### Preconditions: `Mac OS` and `Xcode`

On non-MAC machines, some heuristics are used to check if a directory is most probably a bundle. If that's true, we show the warning to the user.

On MAC machines, the reader performs a strict check with `mdls` (resorting to heuristics if `mdls` is inaccessible). If it deals with a bundle, it checks if `xcresulttool` is available. If it's not, the reader throws an error and shows the user a message with some details on how to get `xcresulttool`.

#### `xcresulttool` API consumption

There are two APIs `xcresulttool` provides: the legacy API and the new API.

The reader tries using the legacy API first as it's more complete and is easier to use. If the legacy API can't be used, it currently throws an error, although this should not happen until Apple removes the legacy API support from a future release of the tool.

There is also an implementation that uses the new API, but it needs to be adjusted to match the legacy-based one (as much as possible). It should also be covered by tests. Once that's done, we should fall back to it. Here is the issue to track the new API-based implementation: #110.

The next Xcode release will be in September 2025 (in 6 months), but we can't know if this will be the one that removes the legacy API support, so it's possible we have even more time.

#### Typed validation

Since it's tricky to create xcresult bundles with invalid structure for testing purposes, the xcresult reader uses a new typed validation module (`reader/src/validation.ts`). That allows shifting the verification from tests to the type system and allows us avoid writing a huge amount of negative tests while still enforcing us to check every input.


packages/reader/src/xcresult/xcresulttool/legacy/cli.ts
```diff
@@ -0,0 +1,46 @@
+import type { Unknown } from "../../../validation.js";
+import { xcresulttool } from "../cli.js";
+import type { LegacyApiOptions } from "./model.js";
+import { getRef } from "./parsing.js";
+import type { XcActionsInvocationRecord, XcReference } from "./xcModel.js";
+
+let legacyRunSucceeded = false;
+let noLegacyApi = false;
+
+let legacyFlagRequired = true;
```
epszaw: I don't like an idea with global variables we mutate in the functions. Could we implement reader as class and isolate this logic inside? At least, you can memoize result in the function.

packages/reader/src/xcresult/xcresulttool/legacy/cli.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+    return undefined;
+  }
+
+  legacyRunSucceeded = true;
+  return result;
+};
+
+export const getRoot = async (xcResultPath: string) =>
+  await xcresulttoolGetLegacy<XcActionsInvocationRecord>(xcResultPath);
```
epszaw: No need to add `await` word here because the function returns promise itself

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
```diff
@@ -0,0 +1,689 @@
...
+  const { xcResultPath } = context;
+  const root = await getRoot(xcResultPath);
+  if (isObject(root)) {
+    const actions = getObjectArray(root.actions);
+    const actionDescriminators = parseActionDiscriminators(actions);
+    const multiTarget = isMultiTarget(actionDescriminators);
+    const multiTestPlan = isMultiTestPlan(actionDescriminators);
+    for (const { actionResult } of actions) {
+      const { destination, testPlan } = actionDescriminators.shift()!;
+      if (isObject(actionResult)) {
```
epszaw: Just check `testsRef` like this `actionResult?.testsRef`. No need to add additional checks like this, in my opinion

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
```diff
@@ -0,0 +1,689 @@
...
+        const { testsRef } = actionResult;
+        const summaries = await getById<XcActionTestPlanRunSummaries>(xcResultPath, testsRef);
+        if (isObject(summaries)) {
+          for (const { testableSummaries } of getObjectArray(summaries.summaries)) {
```
epszaw: Same here:

```js
const { summaries } = await getById<XcActionTestPlanRunSummaries>(xcResultPath, testsRef);

if (summaries) {
  // ...
}
```

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
```diff
@@ -0,0 +1,689 @@
...
+  return actions.map(({ runDestination, testPlanName }) => ({
+    destination: parseDestination(runDestination),
+    testPlan: getString(testPlanName),
+  }));
+};
+
+const isMultiTarget = (discriminators: LegacyActionDiscriminator[]) =>
+  new Set(
+    discriminators
+      .map(({ destination }) => destination)
```
epszaw, line 516: It's possible to reduce iterations here:

```js
discriminators.map(({ destination }) => destination?.name).filter(isDefined)
```

Of course, you can do a single iteration over the items here using `reduce`:

```js
discriminators.reduce((acc, { destination }) => {
  if (destination?.name) {
      return acc.concat(destination.name)
  }
  
  return acc
}, [])
```

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
```diff
@@ -0,0 +1,689 @@
...
+          isTopLevel: explicitTopLevelFlag ?? getBool(isTopLevelFailure),
+        },
+      ] as [string, FailureMapValue])
+    : undefined;
+};
+
+const convertStackTrace = (sourceCodeContext: Unknown<XcSourceCodeContext>) => {
+  if (isObject(sourceCodeContext)) {
+    const { callStack } = sourceCodeContext;
+    return getObjectArray(callStack)
```
epszaw: Reduce iterations here too

epszaw (changes requested):
Unfortunately, most tests fail locally with `xcresulttool version 22608, format version 3.49 (current)` and Xcode 15.4:

image

I guess, we should configure at least two github actions to validate legacy and modern results.

If the modern API works, let's merge the PR and complete legacy support later.

packages/reader/src/xcresult/xcresulttool/legacy/cli.ts
delatrie: I moved the API parsers to classes and did some refactoring. It should be better now (no global state and less context argument traveling up and down the call stack). Those variables now live there:

https://github.com/allure-framework/allure3/blob/d556f604a49c3e3664a5baf14beeabec0ea0d946/packages/reader/src/xcresult/xcresulttool/legacy/index.ts#L101-L103

packages/reader/src/xcresult/xcresulttool/legacy/cli.ts
delatrie: Done

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
delatrie: Hmm... I doubt it. Sure, `reduce` does the job in one go but in terms of complexity it's `O(N^2)` because of allocations `concat` does under the hood (with the only exception when the condition is false for all but a few elements).

`map + filter` on the other way is always `O(2N)`. `O(3N)` at most.

I can do some comparisons later. Maybe I'm mistaken and Node does some fancy optimizations. Although, I don't see where these may come from.

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
delatrie: Can't do exactly that: `actionResult` is `Unknown<XcActionResult>`, which is kinda like plain `unknown`: it can't be destructured without checking first. This is a part of our general approach: use `unknown` to force explicit checks of unreliable user input before using it. We use it in all other readers too.

But your suggestion still makes sense, I just need to wrap objects with `ensureObject` instead. I've done it everywhere where it saves nesting level or otherwise simplifies the code.

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
delatrie: Done!

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
delatrie: It seems my assumptions are correct. Given the following code:

```javascript
const n = parseInt(process.argv[3], 10);

// Input array to work with
const inputArray = new Array(n);
for (let i = 0; i < inputArray.length; i++) {
  inputArray[i] = { index: i };
}

const filterFn = (v) => v % 2 === 0;

// Map + filter alternative
const mapFilter = (input) => input.map(({ index }) => index).filter(filterFn);

// Reduce alternative
const reduce = (input) =>
  input.reduce((acc, { index }) => {
    if (filterFn(index)) {
      return acc.concat(index);
    }
    return acc;
  }, []);

// Get function from the second CLI arg
const fnName = process.argv[2];
const fn = { mapFilter, reduce }[fnName];

if (fn) {
  // Measure the calculation time
  const start = Date.now();

  const res = fn(inputArray);

  const finish = Date.now();

  console.log(fnName, `N=${n}`, `${finish - start}ms,`, res.length, "elements in the result");
}
```

I got the following results:

```
> node ./mapFilterVsReduce.js mapFilter 100000
mapFilter N=100000 3ms, 50000 elements in the result
> node ./mapFilterVsReduce.js reduce 100000
reduce N=100000 4209ms, 50000 elements in the result
```

I also prefer map + filter in terms of readability. No point using `reduce` overall.

delatrie:
@epszaw I fixed the problem with Xcode 15. At least it works on my machine. I would appreciate it if you would check it on yours.

packages/reader/src/xcresult/xcresulttool/legacy/index.ts
epszaw: Ok, no problem

epszaw (approved):
Everything works now:

image

Great job!

PR #85, opened 2025-02-03
Classic updates

todti:
- new version `web-classic` Classic UI based on Allure Awesome using `web-components`
- new plugin `plugin-classic` based on `plugin-awesome`

- the old versions `plugin-classic` and `web-classic` have been moved to `plugin-allure2` and `web-allure2`


allurerc.mjs
```diff
@@ -6,7 +6,7 @@ export default defineConfig({
   plugins: {
     awesome: {
       options: {
-        singleFile: true,
+        singleFile: false,
```
epszaw: Return it back. Should be `true`

allurerc.mjs
```diff
@@ -6,7 +6,14 @@ export default defineConfig({
...
+        reportLanguage: "en",
+        groupBy: ["module", "parentSuite", "suite", "subSuite"],
+      },
+    },
+    classic: {
```
epszaw: Remove classic report, keep only awesome

epszaw (changes requested):

epszaw (changes requested):
`allure2` doesn't work:

image

epszaw (approved):

PR #86, opened 2025-02-04
preserve tree and filters user activity

### LINKED ISSUE #82: Preserve trees leaves/report sections open/close state
Now, if you open some leaves/sections in the report and then navigate to prev/next test result and back, the state resets. We need to keep information about opened UI element (such as tree leaves, specific sections, like test body, before and after hooks) and render their last state.

Dupe of #24

todti:
-closes #82 
- for now,saved filters are stored in LocalStorage too


epszaw (changes requested):
Test body state preserving doesn't work:

https://github.com/user-attachments/assets/c4b02101-4730-4113-a2a2-bc9cb775aa38

packages/web-awesome/src/components/TestResult/TestResultSetup/index.tsx
```diff
@@ -17,18 +18,26 @@ const typeMap = {
...
 export type TestResultSetupProps = {
   setup: AllureAwesomeTestResult["setup"];
+  id?: string;
 };
 
-export const TestResultSetup: FunctionalComponent<TestResultSetupProps> = ({ setup }) => {
-  const [isOpened, setIsOpen] = useState(false);
+export const TestResultSetup: FunctionalComponent<TestResultSetupProps> = ({ setup, id }) => {
+  const teardownId = `${id}-setup`;
+  const isEarlyCollapsed = Boolean(!collapsedTrees.value.has(teardownId));
```
epszaw, line 26: Method `has` already returns `boolean` value, there is no need to have additional conversion here

packages/web-awesome/src/components/TestResult/TestResultSteps/index.tsx
```diff
@@ -15,23 +16,31 @@ const typeMap = {
...
 type StepComponentProps = FunctionalComponent<{
   item?: AllureAwesomeTestStepResult;
   stepIndex?: number;
 }>;
 
-export const TestResultSteps: FunctionalComponent<TestResultStepsProps> = ({ steps }) => {
-  const [isOpened, setIsOpen] = useState(true);
+export const TestResultSteps: FunctionalComponent<TestResultStepsProps> = ({ steps, id }) => {
+  const stepsId = `${id}-steps`;
+  const isEarlyCollapsed = Boolean(!collapsedTrees.value.has(stepsId));
```
epszaw, line 29: Same here. Remove `Boolean` wrapper

packages/web-awesome/src/components/TestResult/TestResultTeardown/index.tsx
```diff
@@ -17,18 +18,27 @@ const typeMap = {
 
 export type TestResultTeardownProps = {
   teardown: AllureAwesomeTestResult["teardown"];
+  id?: string;
```
epszaw: Why `id` is optional? We use it on line `25` without fallback value, so it's `undefined-teardown` by default. As far, as I see, it can't be used without `id`, so let's make the parameter required.

packages/web-awesome/src/components/Tree/Tree.tsx
```diff
@@ -16,9 +17,13 @@ interface TreeProps {
 }
 
 const Tree: FunctionComponent<TreeProps> = ({ tree, statusFilter, root, name, statistic }) => {
-  const [isOpened, setIsOpen] = useState(statistic === undefined || !!statistic.failed || !!statistic.broken);
-  const toggleTree = () => {
+  const isEarlyCollapsed = collapsedTrees.value.has(tree.nodeId);
+  const haveFailedSteps = statistic === undefined || !!statistic?.failed || !!statistic?.broken;
+  const [isOpened, setIsOpen] = useState(() => (isEarlyCollapsed ? !haveFailedSteps : Boolean(haveFailedSteps)));
```
epszaw: Again, why do we need to wrap `haveFailedSteps`? As far as I see, it's `boolean` in any case

epszaw: Don't we need to keep it closed/opened independently of failed steps inside? If the user collapsed the tree, it's ok to keep the tree as is without additional logic.

packages/web-awesome/src/stores/tree.ts
```diff
@@ -15,6 +16,15 @@ export type TreeFiltersState = {
   direction: TreeDirection;
 };
 
+const loadFromLocalStorage = <T>(key: string, defaultValue: T): T => {
```
epszaw: Move to utils, make `defaultValue` optional

epszaw (changes requested):

epszaw (approved):

PR #89, opened 2025-02-06
Feat/add chart component

Viperwow:


packages/web-components/scripts/addPreactAliases.js
```diff
@@ -0,0 +1,42 @@
+import { readFile, writeFile } from "node:fs/promises";
+import path from "node:path";
+import { createRequire } from "node:module";
+
+const require = createRequire(import.meta.url);
+
+const addPreactAliases = async () => {
+  // Determine the tsconfig.json location (assumed to be in the project root)
+  const tsconfigPath = path.join(process.cwd(), "tsconfig.json");
```
epszaw: You can use just `path.resolve("tsconfig.json")` because it returns an absolute path to the file using the process' `cwd` as the base path

packages/web-components/scripts/addPreactAliases.js
```diff
@@ -0,0 +1,42 @@
...
+  let tsconfig;
+
+  try {
+    const content = await readFile(tsconfigPath, "utf8");
+    tsconfig = JSON.parse(content);
+  } catch (err) {
+    process.exit(1);
```
epszaw: The wrapper does nothing because in case of error, the process fails with code 1

packages/web-components/scripts/addPreactAliases.js
```diff
@@ -0,0 +1,42 @@
+import { readFile, writeFile } from "node:fs/promises";
```
epszaw, line 1: Format the file with prettier

packages/web-components/scripts/addPreactAliases.js
```diff
@@ -0,0 +1,42 @@
...
+  // Update the tsconfig paths with trailing separator so TypeScript treats them as directories
+  tsconfig.compilerOptions.paths.react = [ preactCompatDir + path.sep ];
+  tsconfig.compilerOptions.paths["react/jsx-runtime"] = [ preactJsxRuntimeDir + path.sep ];
+  tsconfig.compilerOptions.paths["react-dom"] = [ preactCompatDir + path.sep ];
+  tsconfig.compilerOptions.paths["react-dom/*"] = [ `${preactCompatDir + path.sep  }*` ];
+
+  try {
+    await writeFile(tsconfigPath, JSON.stringify(tsconfig, null, 2), "utf8");
+  } catch (err) {
+    process.exit(1);
```
epszaw: Same here. Remove the wrapper

packages/web-components/scripts/addPreactAliases.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  tsconfig.compilerOptions = tsconfig.compilerOptions || {};
+  tsconfig.compilerOptions.paths = tsconfig.compilerOptions.paths || {};
+
+  const preactCompatFile = require.resolve("preact/compat");
+  const preactJsxRuntimeFile = require.resolve("preact/jsx-runtime");
+
+
+  const preactCompatDir = path.relative(process.cwd(), path.join(path.dirname(preactCompatFile), ".."));
```
epszaw, line 21: Replace `path.join` just by `path.resolve`

epszaw (changes requested):

packages/web-components/scripts/addPreactAliases.js
Viperwow: We need to refer to the package's root, which is why `path.join` was used. `part.resolve` will refer to the `dist` inside a package

packages/web-components/scripts/addPreactAliases.js
Viperwow: Done, no effect on this file

packages/web-components/scripts/addPreactAliases.js
epszaw: Ok, no problem

PR #97, opened 2025-02-13
fix: windows-compatibility-for-tsconfig-generation-script

Viperwow:
- Add slashes conversion to comply with `tsconfig`


epszaw:
#99 already solved the problem

Viperwow:
This is not about #99, this is about `root configs must be focused on node`

PR #108, opened 2025-03-05
Actual expected diff implementation

todti:


packages/web-awesome/src/components/TestResult/TestResultError/TRDiff.tsx
```diff
@@ -0,0 +1,121 @@
...
+const diffFunctions = {
+  chars: diffChars,
+  words: diffWords,
+  lines: diffLines,
+} as const;
+
+type DiffType = keyof typeof diffFunctions;
+type ViewMode = "unified" | "side-by-side";
+
+export const TRDiff = ({ expected, actual }: { expected: string; actual: string }) => {
```
baev: TrDiff

baev (changes requested):

epszaw (changes requested):
When I `actual` or `expected` field contains not a string value the application fails with a runtime error:

image

Make a string conversion before you build diff.

---

Add small gap between trace block and the diff button:

image

epszaw (approved):

PR #109, opened 2025-03-05
Config and store methods for new environments

epszaw:


packages/core-api/src/utils/status.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const statusToPriority = (status: TestStatus | undefined) => {
+  if (!status) {
+    return -1;
+  }
+
+  return StatusByPriority.indexOf(status);
+};
+
+export const getWorstStatus = <T>(
```
delatrie: When called with one argument, this function will gladly eat everything you feed it, making `T` the same as `any`:

```ts
// Compiles with no error
const result = getWorstStatus([1, { foo: "bar" }, [3]]);
```

We can fix it by overloading. This will make the API more specific and expressive:

```ts
export function getWorstStatus(items: readonly TestStatus[]): TestStatus | undefined;
export function getWorstStatus<T>(
  items: readonly T[],
  statusAccessor: (item: T) => TestStatus,
): TestStatus | undefined;
// eslint-disable-next-line prefer-arrow/prefer-arrow-functions
export function getWorstStatus<T>(
  items: readonly T[] | readonly TestStatus[],
  statusAccessor?: (item: T) => TestStatus,
) {
  // ...
}
```

I won't insist on it but having an API like this IMO is better.

packages/core-api/package.json
```diff
@@ -32,6 +32,7 @@
     "@typescript-eslint/eslint-plugin": "^8.0.0",
     "@typescript-eslint/parser": "^8.0.0",
     "@vitest/runner": "^2.1.8",
+    "@vitest/snapshot": "^2.1.8",
```
delatrie, line 35: I wonder why we need snapshots in core-api tests. Are you sure that's not a left-over? Can't see any related imports in the test files.

packages/core/src/api.ts
```diff
@@ -1,27 +1,47 @@
...
   plugins?: PluginInstance[];
   history: HistoryDataPoint[];
-  historyPath: string;
   appendHistory?: boolean;
-  knownIssuesPath: string;
   known?: KnownTestFailure[];
-  qualityGate?: QualityGateConfig;
-  defaultLabels?: DefaultLabelsConfig;
-  // TODO: https://github.com/qameta/allure3/issues/180
+  /**
```
delatrie, line 24: This JSDoc should be moved to `defaultLabels` (currently, it describes `realTime`).

packages/core/src/store/store.ts
```diff
@@ -353,4 +375,21 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
+
+  async testResultsByEnvironment(
+    env: string,
+    options: {
+      includeHidden: boolean;
+    } = { includeHidden: false },
+  ) {
+    const allTrs = await this.allTestResults(options);
+
+    return allTrs.filter((tr) => tr.environment === env);
```
delatrie, line 393: Not a change request, just some thoughts for the future: we should think about introducing some sort of indexing, as iterating over all test results may blow up easily for large reports

packages/plugin-awesome/src/model.ts
```diff
@@ -6,6 +8,7 @@ export type AllureAwesomeOptions = {
   reportLanguage?: "en" | "ru";
   groupBy?: string[];
   layout?: "base" | "split";
+  environments?: Record<string, (tr: TestResult) => boolean>;
```
delatrie: Shouldn't it be `EnvironmentsConfig`?

delatrie (approved):
I've found nothing severe but pls check my comments before merging.

packages/core/src/store/store.ts
epszaw: For sure, but at this moment it's a single way to aggregate test results related data

packages/core/src/api.ts
epszaw: For sure 👍

packages/core-api/src/utils/status.ts
epszaw: Why not. It doesn't matter where do we perform iteration, inside or outside the function. Like your idea

packages/plugin-awesome/src/model.ts
epszaw: Correct, it should

PR #115, opened 2025-03-11
fix hash navigation issues

todti:


packages/web-awesome/src/components/TestResult/TestResultHistory/TestResultHistoryItem.tsx
```diff
@@ -18,7 +18,7 @@ export const TestResultHistoryItem: FunctionalComponent<{
   const formattedDuration = formatDuration(duration as number);
   const { t } = useI18n("controls");
 
-  const navigateUrl = `/${id}`;
+  const navigateUrl = `${id}`;
```
epszaw: Just `id`

packages/web-classic/src/index.tsx
```diff
@@ -41,11 +41,7 @@ const App = () => {
     };
   }, []);
 
-  console.log(route.value.tabName);
-  const ActiveComponent = useMemo(
-    () => tabComponents[route.value.tabName] || (() => <div>nav</div>),
-    [route.value.tabName],
-  );
+  const ActiveComponent = useMemo(() => tabComponents[route.value.tabName] || (() => <div />), [route.value.tabName]);
```
epszaw: Let's return `null` instead of an empty div

epszaw (approved):

PR #116, opened 2025-03-11
Base environments views

epszaw:


packages/web-awesome/src/components/ReportMetadata/index.tsx
```diff
@@ -32,10 +42,44 @@ const Metadata: FunctionalComponent<MetadataProps> = ({ envInfo }) => {
   );
 };
 
+const MetadataVariables: FunctionalComponent<MetadataVariablesProps> = (props) => {
+  const { t } = useI18n("ui");
+  const [isOpened, setIsOpen] = useState(true);
+  const convertedEnvInfo = Object.entries(props.variables).map(([key, value]) => {
+    // return { ...env, value: env.values.join(", ") };
```
todti: remove unused code

packages/web-awesome/src/index.tsx
```diff
@@ -3,18 +3,20 @@ import { Spinner, SvgIcon, allureIcons } from "@allurereport/web-components";
 import "@allurereport/web-components/index.css";
 import clsx from "clsx";
 import { render } from "preact";
-import { useEffect } from "preact/hooks";
+import "preact/debug";
```
todti: debug

todti (changes requested):

packages/web-awesome/src/index.tsx
epszaw: 👍

packages/web-awesome/src/components/ReportMetadata/index.tsx
epszaw: 👍

todti (approved):

packages/core/src/store/store.ts
```diff
@@ -392,4 +401,68 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
+        testResultsByEnv: {},
+      };
+
+      trs.forEach((tr) => {
+        const env = matchEnvironment(this.#environmentsConfig, tr);
+
+        envGroup.testResultsByEnv[env] = tr.id;
+      });
+
+      return acc.concat(envGroup);
```
delatrie: Here, `concat` will copy the accumulator (with all the elements inside) into a new array on each iteration, which turns a linear operation into a quadratic one. The input sequence can be large (up to the total number of test results), so this may affect the startup time quite noticeably.

I would almost always avoid `reduce + concat` pattern in favor of `filter` followed by `map`, `map` followed by `filter`, or `forEach/for..of + if + push`.

I made a quick comparison and here are the results (in seconds):

```js
const input = new Array(parseInt(process.argv.at(-1), 10));
input.fill(0);
const cmd = process.argv.at(-2);

if (cmd === "reduce") {
  input.reduce((acc, inp) => acc.concat({ inp }), []);
} else if (cmd === "map") {
  input.map((acc, inp) => ({ inp }));
} else if (cmd === "forEach") {
  const acc = [];
  input.forEach((inp) => {
    acc.push({ inp });
  });
} else if (cmd === "for") {
  const acc = [];
  for (const inp of input) {
    acc.push({ inp });
  }
}
```

```
N     reduce   map forEach   for
-     ------   --- -------   ---
10000  0.307 0.044   0.043 0.046
20000  0.783 0.049   0.045 0.051
30000  1.816 0.046   0.045 0.059
40000  3.242 0.046   0.047 0.047
50000  5.074 0.048   0.047 0.048
```

packages/core/src/store/store.ts
```diff
@@ -392,4 +401,68 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
+    }, [] as TestEnvGroup[]);
+  }
+
+  async testEnvGroupByTestCaseId(id: string) {
```
delatrie: Did I miss something or this function isn't used anywhere?

delatrie (changes requested):

packages/core/src/store/store.ts
epszaw: Yes, it isn't. Removed the method

packages/core/src/store/store.ts
epszaw: Make sense

delatrie (approved):

lpnueg4:
@epszaw Hi bro, is it possible to enable this feature in python? or How to use this function in python?

PR #131, opened 2025-03-25
Categories awesome implementation

todti:
Add `Categories` and `Suites` toggle 
---
image
image

### DIFF
--- a/packages/plugin-awesome/src/model.ts
+++ b/packages/plugin-awesome/src/model.ts
@@ -1,4 +1,4 @@
-import type { EnvironmentsConfig } from "@allurereport/core-api";
+import type { EnvironmentsConfig, TestStatus } from "@allurereport/core-api";
 
 export type AwesomeOptions = {
   reportName?: string;
@@ -15,6 +15,16 @@ export type AwesomeOptions = {
     name: string;
   };
 };
+
+export interface AwesomeCategory {
+  name: string;
+  description?: string;
+  descriptionHtml?: string;
+  messageRegex?: string;
+  traceRegex?: string;
+  matchedStatuses?: TestStatus[];
+  flaky?: boolean;
+}
 
 export type TemplateManifest = Record<string, string>;
 
--- a/packages/web-awesome/src/components/ReportBody/TreeToggle.tsx
+++ b/packages/web-awesome/src/components/ReportBody/TreeToggle.tsx
@@ -1,0 +1,43 @@
+import { DropdownButton, Menu } from "@allurereport/web-components";
+import { useI18n } from "@/stores";
+import type { TreeSwitcher } from "@/stores/treeSwitcher";
+import { currentTree, setCurrentTree } from "@/stores/treeSwitcher";
+
+export const TreeToggle = () => {
+  const { t } = useI18n("ui");
+
+  const handleSelect = (selectedOption: TreeSwitcher) => {
+    setCurrentTree(selectedOption);
+  };
+  const treeOptions: Record<TreeSwitcher, string> = {
+    suites: t("suites"),
+    categories: t("categories"),
+  };
+
+  return (
+    <Menu
+      size="s"
+      menuTrigger={({ isOpened, onClick }) => (
+        <DropdownButton
+          style="outline"
+          size="m"
+          text={treeOptions[currentTree.value]}
+          isExpanded={isOpened}
+          onClick={onClick}
+        />
+      )}
+    >
+      <Menu.Section>
+        {Object.entries(treeOptions).map(([key, value]) => (
+          <Menu.ItemWithCheckmark
+            onClick={() => handleSelect(key as TreeSwitcher)}
+            key={key}
+            isChecked={currentTree.value === value}
+          >
+            {value}
+          </Menu.ItemWithCheckmark>
+        ))}
+      </Menu.Section>
+    </Menu>
+  );
+};
--- a/packages/web-awesome/src/stores/categories.ts
+++ b/packages/web-awesome/src/stores/categories.ts
@@ -1,0 +1,127 @@
+import { fetchReportJsonData } from "@allurereport/web-commons";
+import { computed, signal } from "@preact/signals";
+import type { AwesomeStatus, AwesomeTree, AwesomeTreeGroup } from "types";
+import type { StoreSignalState } from "@/stores/types";
+import { createRecursiveTree, isRecursiveTreeEmpty } from "@/utils/treeFilters";
+
+export type TreeSortBy = "order" | "duration" | "status" | "alphabet";
+export type TreeDirection = "asc" | "desc";
+export type TreeFilters = "flaky" | "retry" | "new";
+export type TreeFiltersState = {
+  query: string;
+  status: AwesomeStatus;
+  filter: Record<TreeFilters, boolean>;
+  sortBy: TreeSortBy;
+  direction: TreeDirection;
+};
+
+export const categoriesStore = signal<StoreSignalState<AwesomeTree>>({
+  loading: true,
+  error: undefined,
+  data: undefined,
+});
+
+export const noTests = computed(() => !Object.keys(categoriesStore?.value?.data?.leavesById).length);
+
+export const categoriesFiltersStore = signal<TreeFiltersState>({
+  query: "",
+  status: "total",
+  filter: {
+    flaky: false,
+    retry: false,
+    new: false,
+  },
+  sortBy: "order",
+  direction: "asc",
+});
+
+export const filteredCategories = computed(() => {
+  const { root, leavesById, groupsById } = categoriesStore.value.data;
+
+  return createRecursiveTree({
+    group: root as AwesomeTreeGroup,
+    leavesById,
+    groupsById,
+    filterOptions: categoriesFiltersStore.value,
+  });
+});
+
+export const noTestsFound = computed(() => {
+  return isRecursiveTreeEmpty(filteredCategories.value);
+});
+
+export const clearCategoriesFilters = () => {
+  categoriesFiltersStore.value = {
+    query: "",
+    status: "total",
+    filter: {
+      flaky: false,
+      retry: false,
+      new: false,
+    },
+    sortBy: "order",
+    direction: "asc",
+  };
+};
+
+export const setCategoriesQuery = (query: string) => {
+  categoriesFiltersStore.value = {
+    ...categoriesFiltersStore.value,
+    query,
+  };
+};
+
+export const setCategoriesStatus = (status: AwesomeStatus) => {
+  categoriesFiltersStore.value = {
+    ...categoriesFiltersStore.value,
+    status,
+  };
+};
+
+export const setCategoriesSortBy = (sortBy: TreeSortBy) => {
+  categoriesFiltersStore.value = {
+    ...categoriesFiltersStore.value,
+    sortBy,
+  };
+};
+
+export const setCategoriesDirection = (direction: TreeDirection) => {
+  categoriesFiltersStore.value = {
+    ...categoriesFiltersStore.value,
+    direction,
+  };
+};
+
+export const setCategoriesFilter = (filterKey: TreeFilters, value: boolean) => {
+  categoriesFiltersStore.value = {
+    ...categoriesFiltersStore.value,
+    filter: {
+      ...categoriesFiltersStore.value.filter,
+      [filterKey]: value,
+    },
+  };
+};
+
+export const fetchCategoriesData = async () => {
+  categoriesStore.value = {
+    ...categoriesStore.value,
+    loading: true,
+    error: undefined,
+  };
+
+  try {
+    const res = await fetchReportJsonData<AwesomeTree>("widgets/categories.json");
+
+    categoriesStore.value = {
+      data: res,
+      error: undefined,
+      loading: false,
+    };
+  } catch (e) {
+    categoriesStore.value = {
+      ...categoriesStore.value,
+      error: e.message,
+      loading: false,
+    };
+  }
+};
--- a/packages/web-awesome/src/stores/categories.ts
+++ b/packages/web-awesome/src/stores/categories.ts
@@ -1,6 +1,8 @@
 import { fetchReportJsonData } from "@allurereport/web-commons";
 import { computed, signal } from "@preact/signals";
 import type { AwesomeStatus, AwesomeTree, AwesomeTreeGroup } from "types";
+import { treeStore } from "@/stores/tree";
+import { treeFiltersStore } from "@/stores/treeFilters";
 import type { StoreSignalState } from "@/stores/types";
 import { createRecursiveTree, isRecursiveTreeEmpty } from "@/utils/treeFilters";
 
@@ -21,19 +23,7 @@ export const categoriesStore = signal<StoreSignalState<AwesomeTree>>({
   data: undefined,
 });
 
-export const noTests = computed(() => !Object.keys(categoriesStore?.value?.data?.leavesById).length);
-
-export const categoriesFiltersStore = signal<TreeFiltersState>({
-  query: "",
-  status: "total",
-  filter: {
-    flaky: false,
-    retry: false,
-    new: false,
-  },
-  sortBy: "order",
-  direction: "asc",
-});
+export const noTests = computed(() => !Object.keys(categoriesStore?.value?.data?.leavesById || {}).length);
 
 export const filteredCategories = computed(() => {
   const { root, leavesById, groupsById } = categoriesStore.value.data;
@@ -42,7 +32,7 @@ export const filteredCategories = computed(() => {
     group: root as AwesomeTreeGroup,
     leavesById,
     groupsById,
-    filterOptions: categoriesFiltersStore.value,
+    filterOptions: treeFiltersStore.value,
   });
 });
 
@@ -51,7 +41,7 @@ export const noTestsFound = computed(() => {
 });
 
 export const clearCategoriesFilters = () => {
-  categoriesFiltersStore.value = {
+  treeFiltersStore.value = {
     query: "",
     status: "total",
     filter: {
@@ -65,40 +55,16 @@ export const clearCategoriesFilters = () => {
 };
 
 export const setCategoriesQuery = (query: string) => {
-  categoriesFiltersStore.value = {
-    ...categoriesFiltersStore.value,
+  treeFiltersStore.value = {
+    ...treeFiltersStore.value,
     query,
   };
 };
 
 export const setCategoriesStatus = (status: AwesomeStatus) => {
-  categoriesFiltersStore.value = {
-    ...categoriesFiltersStore.value,
+  treeFiltersStore.value = {
+    ...treeFiltersStore.value,
     status,
-  };
-};
-
-export const setCategoriesSortBy = (sortBy: TreeSortBy) => {
-  categoriesFiltersStore.value = {
-    ...categoriesFiltersStore.value,
-    sortBy,
-  };
-};
-
-export const setCategoriesDirection = (direction: TreeDirection) => {
-  categoriesFiltersStore.value = {
-    ...categoriesFiltersStore.value,
-    direction,
-  };
-};
-
-export const setCategoriesFilter = (filterKey: TreeFilters, value: boolean) => {
-  categoriesFiltersStore.value = {
-    ...categoriesFiltersStore.value,
-    filter: {
-      ...categoriesFiltersStore.value.filter,
-      [filterKey]: value,
-    },
   };
 };
 
--- a/packages/plugin-api/src/utils/tree.ts
+++ b/packages/plugin-api/src/utils/tree.ts
@@ -53,6 +53,7 @@ const createTree = <T, L, G>(
     let parentGroups = [root];
 
     for (const layer of itemGroups) {
+      console.log("itemGroups", itemGroups, layer);
       if (layer.length === 0) {
         break;
       }
@@ -181,7 +182,8 @@ export const createTreeByCategories = <T = TestResult, L = DefaultTreeLeaf, G = 
 };
 
 export const byCategories = (item: TestResult): string[][] => {
-  return [item.categories?.map((category: any) => category.name)];
+  console.log("byCategories", item.categories, item.id);
+  return [item.categories?.map((category: TestResult["categories"]) => category.name) || []];
 };
 
 /**
--- a/packages/plugin-awesome/src/categories.ts
+++ b/packages/plugin-awesome/src/categories.ts
@@ -1,0 +1,55 @@
+import type { TestStatus } from "@allurereport/core-api";
+import type { AwesomeCategory } from "./model.js";
+
+const productDefects: AwesomeCategory = {
+  name: "Product defects",
+  matchedStatuses: ["failed"],
+};
+
+const testDefects: AwesomeCategory = {
+  name: "Test defects",
+  matchedStatuses: ["broken"],
+};
+
+export const matchCategories = (
+  categories: AwesomeCategory[],
+  result: { message?: string; trace?: string; status: TestStatus; flaky: boolean },
+) => {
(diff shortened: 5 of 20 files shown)


packages/web-awesome/src/components/SectionTabs/index.tsx
epszaw, line 1: Delete an empty file

packages/web-awesome/src/stores/categories.ts
```diff
@@ -0,0 +1,36 @@
+import type { TestCategories } from "@allurereport/core-api";
+import { fetchReportJsonData } from "@allurereport/web-commons";
+import { computed, signal } from "@preact/signals";
+import type { StoreSignalState } from "@/stores/types";
+
+export const categoriesStore = signal<StoreSignalState<TestCategories>>({
```
epszaw, line 6: Use `LoadableStore` from `web-commons` package here

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,87 @@
...
+  store: CategoriesStore;
+  activeNodeId?: string;
+  depth?: number;
+};
+
+export type ErrorCategoriesMode = "inclusive" | "first";
+
+export type ErrorCategoriesConfig = {
+  mode: ErrorCategoriesMode;
+  defaultCategories: "on" | "off";
```
epszaw: Better to have boolean flag instead

epszaw: Also, it seems the parameter doesn't work property.
Independently of the given value, I always default categories in the generated report.

When nothing is matched:

image

When one category is matched:

image

Here's the config:

```
{
    mode: "inclusive",
    rules: [
      {
        name: "Example",
        matcher: ({ trace }) => {
          return /^Error:.+$/.test(trace)
        }
      }
    ]
}
```

I think, we should generate default categories only when:

1. The flag is truthy
2. There's no matched categories (what do you think, @delatrie?)

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,87 @@
...
+  rules: ErrorCategoryInput[];
```
epszaw: Better to rename `ErrorCategoryInput` to `ErrorCategoryRule`, or `ErrorCategoryConfig`, there's no correlation between `input` and `rules` now.

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export type ErrorCategoryNormalized = {
+  name: string;
+  matchers: Matcher[];
+  applyTags?: string[];
+  group?: string;
+  index: number;
+};
+
+export type CategoryNodeProps = {
```
epszaw, line 113: It seems you're using this type and `CategoriesStore` one only in awesome, so let's transfer the types to the report's package because they don't look as common ones

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,87 @@
...
+  duration?: number;
+};
+
+export type ObjectMatcher = {
+  statuses?: TestStatus[];
+  labels?: Record<string, string | RegExp>;
+  message?: string | RegExp;
+  trace?: string | RegExp;
+  flaky?: boolean;
+  [k: string]: unknown;
```
epszaw: Do we need this? I think we have to include only the known properties to the declarative matcher

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export type PredicateMatcher = (d: ErrorMatchingData) => boolean;
+
+export type Matcher = ObjectMatcher | PredicateMatcher;
+
+export type ErrorCategoryInput = {
+  name: string;
+  matchers?: Matcher[];
+  matcher?: Matcher;
```
epszaw: Keep just `matchers`, `matcher` looks like a duplicate of the field above.

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,87 @@
...
+  matchedStatuses?: TestStatus[];
+  messageRegex?: string;
+  traceRegex?: string;
+  flaky?: boolean;
+  applyTags?: string[];
```
epszaw: `tagsToApply`?

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,87 @@
...
+  group?: string;
+};
+
+export type CategoriesStore = {
+  roots: string[];
+  nodes: Record<string, CategoryNode>;
+};
+
+export type ErrorCategoryNormalized = {
```
epszaw: Use `Pick<ErrorCategoryInput, ...>` because the most of the fields can be re-used

packages/web-awesome/src/components/ReportBody/TreeToggle.tsx
```diff
@@ -0,0 +1,43 @@
+import { DropdownButton, Menu } from "@allurereport/web-components";
+import { useI18n } from "@/stores";
+import type { TreeSwitcher } from "@/stores/treeSwitcher";
+import { currentTree, setCurrentTree } from "@/stores/treeSwitcher";
+
+export const TreeToggle = () => {
```
epszaw: The component is unused. Did you decide to go with tabs in the final result?

epszaw: If you did – delete the component

packages/plugin-awesome/src/errorCategories/normalize.ts
```diff
@@ -0,0 +1,122 @@
...
+      matchers = [m];
+    } else {
+      throw new Error(`errorCategories.rules[${index}] must define matcher/matchers or compatibility keys`);
+    }
+
+    if (!matchers.length) {
+      throw new Error(`errorCategories.rules[${index}].matchers is empty`);
+    }
+
+    for (let i = 0; i < matchers.length; i++) {
```
epszaw: Why can't we use `forEach` there?

epszaw (changes requested):

packages/web-awesome/src/stores/categories.ts
todti: will be in the next PR's

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,296 @@
...
+      matchers = [...rule.matchers];
+    } else {
+      matchers = [rule.matchers];
+    }
+  } else if (compatKeysUsed) {
+    const compatMatcher: ObjectMatcher = {};
+    if (rule.matchedStatuses) {
+      compatMatcher.statuses = rule.matchedStatuses;
+    }
+    if (rule.messageRegex) {
```
delatrie: This evaluates to `false` for an empty string, which is a valid pattern that means "the message is any string."

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,296 @@
...
+      compatMatcher.message = rule.messageRegex;
+    }
+    if (rule.traceRegex) {
```
delatrie: The same as above

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,296 @@
...
+      compatMatcher.trace = rule.traceRegex;
+    }
+    if (rule.flaky !== undefined) {
+      compatMatcher.flaky = rule.flaky;
+    }
+    matchers = [compatMatcher];
+  }
+
+  if (matchers.length === 0) {
+    throw new Error(`errorCategories[${index}] must define matchers or compatibility keys`);
```
delatrie: Let's not advertise compatibility keys here. If nothing is defined, we should urge people to use the new-style matchers.

packages/plugin-awesome/src/model.ts
```diff
@@ -20,6 +20,7 @@ export type AwesomeOptions = {
   defaultSection?: string;
   publish?: boolean;
   appendTitlePath?: boolean;
+  categories?: ErrorCategoriesConfig;
```
delatrie: Categories are a global feature, not Awesome-specific. Kinda like Environments. It should be a top-level key in the config.

packages/web-awesome/src/components/Categories/MessageTreeItem/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const sanitizedMessage = ansiToHTML(node.name ?? "", { fg: "var(--on-text-primary)", colors: {} });
+  const stickyStyle = createCategoriesStickyStyle(depth);
+
+  return (
+    <div className={clsx(styles["tree-content"], styles["tree-item-message"])} id={nodeId}>
+      <div
+        className={styles["tree-item-message-container"]}
+        data-tree-header
+      >
+        <ArrowButton isOpened={isOpened}  className={styles["tree-item-message-arrow"]} />
```
delatrie: This button does not work:

https://github.com/user-attachments/assets/bb20d5e9-cdda-4ca6-9c4a-8c8d72b237bc

packages/web-awesome/src/components/Categories/MessageTreeItem/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <div className={clsx(styles["tree-item-message-card"], status && styles[`message-status-${status}`])} style={stickyStyle}>
+          <div className={styles["tree-item-message-header"]} onClick={onToggle}>
+            <Code size="s">
+              {/* eslint-disable-next-line react/no-danger */}
+              <pre dangerouslySetInnerHTML={{ __html: sanitizedMessage }} />
```
delatrie: If the message is long, it breaks the markup:

Image

Since the mssages can be quite long, let's think how to make them easier to work with. For example, we can initially limit them with one line (adding `...` if it doesn't fit), with the ability to expand the full message by clicking a button or the message itself.

delatrie: If we collapse multiple messages, they appear very close to each other creating an impression that we're dealing with a single multiline message, which is not the case:

Image

How about introducing a tiny margin between them?

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,296 @@
+import type { Statistic } from "./aggregate.js";
+import type { TestLabel } from "./metadata.js";
+import type { TestResult, TestStatus, TestStatusTransition } from "./model.js";
+
+export type TestCategories = {
+  roots: string[];
+  nodes: Record<string, CategoryNode>;
+};
+
+export type ErrorMatchingData = {
```
delatrie: `ErrorMatchingData` should also include `transition` and `environment`.

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,296 @@
...
+  status: TestStatus;
+  labels: readonly TestLabel[];
+  message?: string;
+  trace?: string;
+  flaky: boolean;
+  duration?: number;
+};
+
+export type ObjectMatcher = {
```
delatrie, line 56: `ObjectMatcher` should also include `transitions?: readonly TestStatusTransition[]` and `environments?: readonly string[]`.

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,296 @@
...
+  flaky?: boolean;
+};
+
+export type PredicateMatcher = (d: ErrorMatchingData) => boolean;
+
+export type Matcher = ObjectMatcher | PredicateMatcher;
+
+export type CategoryMatcher = Matcher | readonly Matcher[];
+
+export type CategoryGroupBuiltInSelector = "flaky" | "owner" | "severity" | "transition";
```
delatrie: It should also include `status` and `environment`. Let's also add `layer` as it's a popular choice for grouping.

The `environment` selector does what currently `groupByEnvironment` does but is more flexible in terms of what level the grouping works on.

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,296 @@
...
+
+export type CategoryGroupSelector = CategoryGroupBuiltInSelector | CategoryGroupCustomSelector;
+
+export type ErrorCategoryRule = {
+  name: string;
+  matchers?: CategoryMatcher;
+  groupBy?: readonly CategoryGroupSelector[];
+  groupByMessage?: boolean;
+  groupByEnvironment?: boolean;
+  groupByHistoryId?: boolean;
```
delatrie: These two should be replaced with a single switch. Let's name it `groupEnvironments` or `groupTestEnvironments` as it better reflects what it does.

From the [spec](https://github.com/allure-framework/allure-org/issues/212#issuecomment-3616686846):

> If groupEnvironments is true, a new level is added that groups together results of the same test in different environments. This level is always the last one: it's added after the levels denoted by groupBy and groupByMessage:
> 
>  ```
>   - category
>     - groups from `groupBy`
>       - message
>         - test
>           - test result from environment <- this is a leaf
> ```
> 
> The default value for groupEnvironments is calculated according to these rules:
> 
>   - The default value is true if both conditions hold:
>       1. More than one environment exists
>       2. No environment selector is present in groupBy.
>   - Otherwise, the default value is false.
> 
> Additionally, if a single environment is selected in the dropdown, this grouping must be ignored.

Basically, when set to `true`, it does what currently a combination of `groupByEnvironment: true` and `groupByHistoryId: true` does.

Given we now have `environment` in the build-in group specifiers, there is no need in the second flag.

packages/web-awesome/src/components/Categories/GroupTreeItem/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+        isOpened={isOpened}
+        categoryTitle={headerTitle}
+        toggleTree={onToggle}
+        data-tree-header
+        style={stickyStyle}
+        reportStatistic={reportStatistic}
+        statistic={node.statistic}
+        statusFilter={"total"}
+      />
+      {isOpened && children}
```
delatrie, line 73: We should provide the default ordering for the groups. I've updated the [spec](https://github.com/allure-framework/allure-org/issues/212#issuecomment-3616686846):

> Each grouping has the default sorting order:
> 
> `status`: the canonical order (failed, broken, passed, skipped, unknown).
> `severity`: from blocker to trivial, no severity at the end.
> `owner`: alphabetical, no owner at the end.
> `layer`: alphabetical, no layer at the end.
> `message`: by the number of test results.
> `transition`: the canonical order (regressed, malfunctioned, new, fixed).
> `flaky`: true, then false.
> `environment`: the same order as in the config. The default environment comes last.
> Custom labels: alphabetical, the missing value is at the end.

Note, all missing values must be at the bottom. Currently, they are sorted together with real values:

Image

packages/web-awesome/src/components/Categories/SeverityTreeItem/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+      subtreeToggle={subtreeToggle}
+      reportStatistic={reportStatistic}
+      className={styles["tree-item-severity"]}
+      title={
+        <div className={styles["tree-item-severity-title"]}>
+          <TrSeverity severity={severityValue} size="m" />
+        </div>
+      }
+    >
+      {children}
```
delatrie, line 47: Pls, check the order. Currently, `minor` goes before `normal`. It must be vice versa.

packages/web-awesome/src/components/Categories/LabelTreeItem/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      isOpened={isOpened}
+      onToggle={onToggle}
+      depth={depth}
+      subtreeToggle={subtreeToggle}
+      reportStatistic={reportStatistic}
+      className={styles["tree-item-label"]}
+      title={
+        <div className={styles["tree-item-label-row"]}>
+          <Text type={"ui"} size="m" className={styles["tree-item-label-key"]}>
+            {node.key}
```
delatrie: How about closing the gap between the label name and its value? It seems pretty big to me.

Image

packages/core-api/src/categories.ts
delatrie: Let's translate the built-in group tags when displaying it in the tree: `owner`, `layer`, `environment`, `flaky`, `status`, and `transition`. Additionaly, let's translate status and transition values.
Also, let's use something more user-friendly instead of code-specific `true` and `false` for `flaky`. For example: `Flaky: Yes` and `Flaky: No`, or even simpler: `Flaky` and `Not flaky`. And translate that as well.

packages/web-awesome/src/components/Categories/CategoryTreeItem/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+        subtreeToggle={subtreeToggle}
+        reportStatistic={reportStatsStore.value.data}
+      >
+        {renderChildren()}
+      </GroupTreeItem>
+    );
+  }
+  if (node.type === "tr") {
+    return (
+      <div className={styles["tree-item"]} id={nodeId}>
```
delatrie, line 297: The titles of icons (like flaky, transition, retries) are not visible. On the screenshot below, I've hovered my mouse over the flaky icon:

Image

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,473 @@
...
+    status: tr.status,
+    labels,
+    message,
+    trace,
+    flaky: tr.flaky,
+    duration: tr.duration,
+  };
+};
+
+export const buildEnvironmentSortOrder = (environmentNames: string[], defaultEnvironmentName: string) => {
```
delatrie, line 359: It works for explicit groups (`groupBy: ["environment"]`).
It should also work for `groupByEnvironment`:

Image

packages/core-api/src/categories.ts
delatrie: We've ended up with this:

  - no flag (default cats are always appended)
  - hide if nothing matches (applies to all cats)
  - support `hide: true` to preclude test results from the page

packages/web-awesome/src/components/Categories/MessageTreeItem/index.tsx
delatrie: The overflow issue is fixed. Now I see ellipsis for long messages. Should we allow viewing the full message in-place somehow?

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,473 @@
...
+    }
+  }
+
+  if (m.trace !== undefined) {
+    const re = toRegExp(m.trace);
+    if (!re.test(d.trace ?? "")) {
+      return false;
+    }
+  }
+
```
delatrie, line 305: Also check against `transitions` and `environments`

packages/web-awesome/src/components/TestResult/TrSeverity/index.tsx
```diff
@@ -10,17 +10,26 @@ const icons: Record<string, string> = {
...
+}) => {
   const { t } = useI18n("severity");
   const statusClass = clsx(styles[`severity-${severity}`]);
 
   return (
     <div className={styles["test-result-severity"]}>
-      <SvgIcon className={statusClass} id={icons[severity]} />
-      <Text size={"s"} bold className={styles["test-result-severity-text"]}>
-        {capitalize(t(severity))}
+      <SvgIcon className={statusClass} id={icons[severity]} size={size} />
```
delatrie, line 30: Severity icons look pretty cool in the categories tree. WDYT about adding icons to status and, posibly, transition nodes? We already have icons for them.

packages/web-awesome/src/components/Categories/CategoryTreeItem/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+        nextCollapsed.add(id);
+      }
+    });
+
+    collapsedTrees.value = nextCollapsed;
+  };
+
+  const onToggleSubtree = (event: MouseEvent) => {
+    event.stopPropagation();
+    if (hasOnlyLeafResults) {
```
delatrie, line 116: Currently, this toggle has three states:

  - node collapsed: the toggle is shown as `v` and does exactly what the row itself does when clicked
  - node expanded but all its children are collapsed: the toggle is shown as `︾` and has a unique behavior: it expands all the children.
  - node expanded and at least one child is expanded: the toggle is shown as `︽`. It behaves mostly like the row itself (collapses the node) but has a unique extra behavior: it makes the children collapsed when the node is later expanded.

I like it. But I often found myself clicking it twice in a row to display the node's children collapsed. Maybe we should change it a little?

  - node collapsed - as before
  - node expanded but all its children are collapsed - as before
  - node expanded and a child is expanded - collapse the children (but not the node)

It matches what VSCode does (and many other IDEs, I assume). Collapsing the node can always be done by clicking on the row.

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,473 @@
...
+  | "environment"
+  | "layer";
+
+export type CategoryGroupCustomSelector = {
+  label: string;
+};
+
+export type CategoryGroupSelector = CategoryGroupBuiltInSelector | CategoryGroupCustomSelector;
+
+export type ErrorCategoryRule = {
```
delatrie: Let's be consistent with the naming and change `ErrorCategory*` to `Category*`

delatrie (changes requested):
There are some unaddressed comments from the previous review. Please check them

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,499 @@
...
+    matchers: { statuses: ["broken"] },
+  },
+];
+
+export type TestCategories = {
+  roots: string[];
+  nodes: Record<string, CategoryNode>;
+};
+
+export type ErrorMatchingData = {
```
delatrie: rename to `CategoryMatchingData`?

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,499 @@
...
+  | "environment"
+  | "layer";
+
+export type CategoryGroupCustomSelector = {
+  label: string;
+};
+
+export type CategoryGroupSelector = CategoryGroupBuiltInSelector | CategoryGroupCustomSelector;
+
+export type ErrorCategoryRule = {
```
delatrie: rename to `CategoryRule`?

packages/core-api/src/categories.ts
```diff
@@ -0,0 +1,499 @@
...
+  traceRegex?: string;
+  flaky?: boolean;
+};
+
+export type CategoriesStore = {
+  roots: string[];
+  nodes: Record<string, CategoryNode>;
+};
+
+export interface ErrorCategoryNorm extends Pick<ErrorCategoryRule, "name" | "expand" | "hide" | "groupEnvironments"> {
```
delatrie: Rename to `NormalizedCategory` or, even better, `CategoryDefinition`?

packages/core/src/report.ts
```diff
@@ -715,6 +719,7 @@ export class AllureReport {
         reportUrl: this.reportUrl,
         output: this.#output,
         ci: this.#ci,
+        categories: this.#categories,
```
delatrie, line 729: Let's move the normalization from plugins to here because, currently, every plugin that uses categories must normalize them anyway.

The normalization is not computationally expensive, so it should be no problem to invoke it even if the currently configured list of plugins doesn't include one that uses them.

packages/web-awesome/src/components/TestResult/TrSeverity/index.tsx
delatrie: Transition nodes look good IMO 👍 
As for status nodes, how about styling them similarly to severities?
Compare:

Image
Image

delatrie (changes requested):
There are some tiny naming inconsistencies and styling/optimization suggestions. Please check the commands. 

Also, the following comments are not addressed:

- [x] https://github.com/allure-framework/allure3/pull/131#discussion_r2826546974
- [x] https://github.com/allure-framework/allure3/pull/131#discussion_r2797448883
- [x] https://github.com/allure-framework/allure3/pull/131#discussion_r2826786706
- [x] https://github.com/allure-framework/allure3/pull/131#discussion_r2798042554

Let's at least fix the matching against `transitions` and `environments`.

Issue #130, opened 2025-03-25
Impossible to see anything in text with ANSI colors


epszaw:
**Describe the bug**

![Image](https://github.com/user-attachments/assets/e231da0f-d5d2-44ad-be22-8c68a43c6f73)

**To Reproduce**
1. Generate Awesome report for the test result file ([45026f6d-ce92-4df1-ba73-b9b784abeb0c-result.json](https://github.com/user-attachments/files/19453782/45026f6d-ce92-4df1-ba73-b9b784abeb0c-result.json))
2. Open the test result in the generated report
3. Enable light theme

PR #134, opened 2025-04-01
awesome:  actual/expected fixes

### LINKED ISSUE #129: Broken layout for error diff viewer
**To Reproduce**
1. Generate some data in https://json-generator.com
2. Use `expect().toMatchObject()` from Vitest, to compare the data
3. Generate Awesome report based on the test results

**Screenshots**
Image

todti:
closes:
#129
#130

### DIFF
--- a/packages/web-components/src/components/Modal/styles.scss
+++ b/packages/web-components/src/components/Modal/styles.scss
@@ -27,6 +27,7 @@
   min-width: 320px;
   min-height: 320px;
   max-height: 80vh;
+  height: 80vh;
   display: flex;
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/Modal/styles.scss
```diff
@@ -27,6 +27,7 @@
   min-width: 320px;
   min-height: 320px;
   max-height: 80vh;
```
epszaw, line 29: Guess, we don't need `max-height` because you set `height` value below

epszaw (approved):

PR #140, opened 2025-04-03
Add option to exclude skipped from report chart

davidashman:


epszaw:
Hello, @davidashman! Thank you for the contribution. [Allure Awesome config provides an ability to filter tests](https://github.com/allure-framework/allure3/blob/main/packages/plugin-awesome/src/model.ts#L17), so you can do that without additional functionality:

```js

export default defineConfig({
  name: "Allure Report 3",
  output: "./out/allure-report",
  plugins: {
    awesome: {
      options: {
        filter: ({ status }) => status !== "skipped",
      },
    },
  },
});
```

The feature is not documented so far (and I even doubt that we released the feature), but I'm going to release next version week.

davidashman:
Thanks for the info!  Glad this is on the horizon.  I looked at the code and what I see is that the filter only applies to the plugin "info" method, but not the generated report.  Am I reading it wrong?  Additionally, I would love to see the ability to separate the general filtering of test results in the report AND the ability to specifically filter in the chart.  The reason is this - it's highly valuable for my developers to SEE the skipped tests, but they don't actually count against the PASS rate in the chart.  They are just skipped for some specific reason.  What I'd like to see is a pass rate that shows how many of the tests we ran actually passed and how many failed.

thanks
dave

epszaw:
Hm, you're right, it's definitely a bug. We'll fix it, thank you for the investigation!

epszaw:
The PR with working filters was merged, the feature will be available in the next release. Closing the PR because the problem was solved. Thank you!

PR #143, opened 2025-04-08
awesome: show broken/failed steps inside success step

### LINKED ISSUE #136: Reveal failed/broken steps inside passed ones
We see test steps as a runtime logging mechanism, which means we prefer steps to follow the natural execution flow of the program. That also means a passed step may contain broken/failed children.

A couple of examples:

  - Code-level retries, fallbacks, and other error-recovery strategies
  - Soft assertions

Currently, such steps are hidden inside a passed one (see the screenshots), and there is no way to figure out that something is not OK inside except by checking manually.

**Steps are collapsed:**

Image

**When we expand the step manually:**

Image

We should indicate somehow that a step has something failed/broken inside.

todti:
fixed #136

### DIFF
--- a/packages/web-awesome/src/components/TestResult/TrSteps/TrStep.tsx
+++ b/packages/web-awesome/src/components/TestResult/TrSteps/TrStep.tsx
@@ -1,4 +1,4 @@
-import type { DefaultTestStepResult } from "@allurereport/core-api";
+import type { DefaultTestStepResult, TestStepResult } from "@allurereport/core-api";
 import { Code, Text, TreeItemIcon, allureIcons } from "@allurereport/web-components";
 import type { FunctionComponent } from "preact";
 import { useState } from "preact/hooks";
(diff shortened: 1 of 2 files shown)


packages/web-awesome/src/components/TestResult/TrSteps/TrStep.tsx
```diff
@@ -44,12 +44,21 @@ export const TrStepsContent = (props: { item: DefaultTestStepResult }) => {
   );
 };
 
+const hasFailedStep = (step: TestStepResult): boolean => {
+  if (step.type !== "step") {
+    return false;
+  }
+
+  return step.status !== "passed" || step.steps.some(hasFailedStep);
```
delatrie, line 52: Should we also collapse parents of skipped steps by default?

todti: skipped steps are collapsed by default, unless they were previously opened

delatrie: I mean its parents, not the step itself. For example:

```ts
test("test auto-expansion", async () => {
  await step("a parent with a skipped sub-step", async () => {
    logStep("skipped", Status.SKIPPED);
  });
  await step("a parent with a passed sub-step", async () => {
    logStep("skipped");
  });
});
```

Currently, the first step is expanded by default, while the second one is hidden:

image

epszaw (approved):

PR #144, opened 2025-04-08
wip:awesome: show test errors in tr's tree

todti:
image

What if we show test errors directly in the test result tree?

### DIFF
--- a/packages/plugin-awesome/src/generators.ts
+++ b/packages/plugin-awesome/src/generators.ts
@@ -178,7 +178,7 @@ export const generateTree = async (
   const tree = createTreeByLabels<AwesomeTestResult, AwesomeTreeLeaf, AwesomeTreeGroup>(
     visibleTests,
     labels,
-    ({ id, name, status, duration, flaky, start, retries }) => {
+    ({ id, name, status, duration, flaky, start, retries ,error}) => {
       return {
         nodeId: id,
         retry: !!retries?.length,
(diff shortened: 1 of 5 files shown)


packages/web-components/src/components/Tree/TreeItem.tsx
```diff
@@ -13,6 +14,8 @@ interface TreeItemProps {
   groupOrder: number;
   marked?: boolean;
   navigateTo: (id: string) => void;
+  error?: TestError;
+  ErrorComponent: FunctionComponent<TestError & { status?: TestStatus }>;
```
epszaw, line 18: Let's use render props function (like `renderError`) instead of component itself

packages/plugin-awesome/src/generators.ts
```diff
@@ -178,7 +178,7 @@ export const generateTree = async (
   const tree = createTreeByLabels<AwesomeTestResult, AwesomeTreeLeaf, AwesomeTreeGroup>(
     visibleTests,
     labels,
-    ({ id, name, status, duration, flaky, start, retries }) => {
+    ({ id, name, status, duration, flaky, start, retries ,error}) => {
```
epszaw, line 181: Run `prettier`

epszaw (changes requested):
Solve conflicts in the branch

epszaw:
We added an error steps feature that solves the problem

PR #148, opened 2025-04-12
Add ability to have different filters for the report data

davidashman:
Gives the ability to filter the chart, statistic and report data with different filters.


baev:
Hi @davidashman 

Thanks for the contribution! Please elaborate a little and provide as much context as possible on the issue you're trying to solve. Thanks

PR #151, opened 2025-04-17
Allure remote history

epszaw:
The PR provides an ability to use remote history served by Allure Service.

New config field `allureService` sets-up the project:

```diff
import { defineConfig } from "allure";

export default defineConfig({
  name: "Allure Report 3",
  output: "./out/allure-report",
  plugins: {
    awesome: {
      options: {
        singleFile: false,
        reportLanguage: "en",
        reportName: "Allure 3 Report",
      },
    },
  },
+  allureService: {
+    project: "allure3",
+    publish: true
+  }
});
```

## How to start using remote history

1. Use `allure login` command to sign in into the Allure Service

image

2. Once you're signed in, use `allure whoami` command to check, that you're using a correct account

image

4. Run `allure init` command in the project root

image

5. Insert code snipped from stdout to the config
6. Run tests, the report's history now be uploaded and downloaded from the Allure Service automatically

## Additional CLI commands to check

- `allure projects` - renders projects list and allows picking one to get a code snippet to insert to the config
- `allure project-create` – creates a new project
- `allure project-delete` – deletes a project
- `allure logout` – revokes current session

## How to test the feature locally

1. Start Allure Service locally
2. Add the service url to the config:

```diff
import { defineConfig } from "allure";

export default defineConfig({
  name: "Allure Report 3",
  output: "./out/allure-report",
  plugins: {
    awesome: {
      options: {
        singleFile: false,
        reportLanguage: "en",
        reportName: "Allure 3 Report",
      },
    },
  },
  allureService: {
    project: "allure3",
+   url: "http://localhost:12345" // url your instance running on
  }
});
```
3. Complete initial flow above


packages/cli/src/commands/projects/list.ts
```diff
@@ -0,0 +1,54 @@
...
+import { createCommand } from "../../utils/commands.js";
+
+export const ProjectsListCommandAction = async () => {
+  const config = await readConfig();
+  const service = new AllureService(config?.allureService);
+  const projects = await service.projects();
+
+  if (projects.length === 0) {
+    // eslint-disable-next-line no-console
+    console.info(yellow("No projects found. Create a new one with `allure create-project` command"));
```
delatrie:
```suggestion
    console.info(yellow("No projects found. Create a new one with `allure project-create` command"));
```

packages/cli/src/commands/logout.ts
```diff
@@ -0,0 +1,20 @@
...
+  const service = new AllureService(config?.allureService);
+
+  await service.logout();
+  // eslint-disable-next-line no-console
+  console.info(green("Logged out"));
+};
+
+export const LogoutCommand = createCommand({
+  name: "logout",
+  description: "Logouts to the Allure Service",
```
delatrie:
```suggestion
  description: "Logs out of the Allure Service",
```

packages/cli/src/commands/login.ts
```diff
@@ -0,0 +1,20 @@
...
+  const service = new AllureService(config?.allureService);
+
+  await service.login();
+  // eslint-disable-next-line no-console
+  console.info(green("Logged in"));
+};
+
+export const LoginCommand = createCommand({
+  name: "login",
+  description: "Logins to the Allure Service",
```
delatrie:
```suggestion
  description: "Logs in to the Allure Service",
```

packages/core/src/history.ts
```diff
@@ -24,7 +24,10 @@ const createHistoryItems = (testResults: TestResult[]) => {
     })
     .reduce(
       (previousValue, currentValue) => {
-        previousValue[currentValue.historyId] = currentValue;
+        previousValue[currentValue.historyId] = {
```
delatrie: This confused me at first. How about changing the names from `previousValue` and `currentValue` to something like `aggregatedValue/historyMap/...` and `testResult`?

packages/core/src/report.ts
```diff
@@ -82,6 +94,16 @@ export class AllureReport {
 
     // TODO: where should we execute quality gate?
     this.#qualityGate = new QualityGate(qualityGate);
+
+    if (allureServiceConfig) {
+      this.#allureService = new AllureService(allureServiceConfig);
+    }
```
delatrie: AllureService is already created at line 68

packages/core/src/report.ts
```diff
@@ -82,6 +94,16 @@ export class AllureReport {
...
+
+    if (allureServiceConfig) {
+      this.#history = new AllureRemoteHistory(this.#allureService!);
```
delatrie: nit: if we check `this#allureService` instead, no type assertion is needed

packages/plugin-api/src/config.ts
```diff
@@ -39,6 +39,12 @@ export interface Config {
   plugins?: Record<string, PluginDescriptor>;
   qualityGate?: QualityGateConfig;
   appendHistory?: boolean;
+  allureService?: {
+    url?: string;
+    project?: string;
+    accessToken?: string;
```
delatrie, line 45: With `accessToken`, the whole config becomes unpublishable. IMO, that creates too much burden on users. Maybe it'd be better to use a separate file for such a secret? We may also support env vars to ease CI jobs a bit.

For example:

```js
const readAccessToken = async (config: Config) => {
  return (
    process.env.ALLURE_SERVICE_ACCESS_TOKEN ??
    (await readFile(config.allureService?.accessTokenPath ?? DEFAULT_ACCESS_TOKEN_PATH, { encoding: "utf-8" }))
  );
};
```

packages/plugin-awesome/src/generators.ts
```diff
@@ -50,6 +50,10 @@ const template = `<!DOCTYPE html>
 </head>
 <body>
     <div id="app"></div>
+    <script>
+      document.cookie = "reportUuid=;path=/;expires=Thu, 01 Jan 1970 00:00:01 GMT";
+      document.cookie = "reportUuid={{ reportUuid }};path = /;max-age=31536000";
```
delatrie, line 55: The scope of cookies is the browser + the host. If multiple reports hosted in Allure Service are opened in separate tabs, some unexpected behavior can be observed when interacting with all but the last report. In particular, the test pages won't open, resulting in a 404 error.

Since we have some other isolation issues (I will create an issue soon), I suggest keeping this implementation and fixing the issue in the follow-up work.

delatrie, line 55: Also, for some reason, when an Allure Awesome report published in Allure Service is opened for the first time, a 505 error is thrown on the service side (an empty page is rendered).

To reproduce, delete all cookies for localhost:8889 and open/reload a report tab.

packages/service/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+- Learn more about Allure Report at https://allurereport.org
+- 📚 [Documentation](https://allurereport.org/docs/) – discover official documentation for Allure Report
+- ❓ [Questions and Support](https://github.com/orgs/allure-framework/discussions/categories/questions-support) – get help from the team and community
+- 📢 [Official announcements](https://github.com/orgs/allure-framework/discussions/categories/announcements) – be in touch with the latest updates
+- 💬 [General Discussion ](https://github.com/orgs/allure-framework/discussions/categories/general-discussion) – engage in casual conversations, share insights and ideas with the community
+
+---
+
+TODO:
```
delatrie, line 13: Should a proper README be added as a part of this PR or later?

packages/service/src/utils/http.ts
```diff
@@ -0,0 +1,57 @@
...
+  const sendRequest =
+    (method: "get" | "delete") =>
+    async (endpoint: string, params?: Record<string, any>, options?: AxiosRequestConfig) => {
+      const actualAccessToken = accessToken || (await readAccessToken());
+      const headers = {
+        ...(options?.headers ?? {}),
+      };
+
+      if (actualAccessToken) {
+        headers.Authorization = actualAccessToken;
```
delatrie: I would stick to [the standard syntax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Authorization#syntax) of the `Authorization` header. In our case, the scheme should be `Bearer`:
```js
headers.Authorization = `Bearer ${actualAccessToken}`;
```

packages/service/test/index.test.ts
```diff
@@ -0,0 +1,3 @@
+import { it } from "vitest";
+
+it.todo("should generate files", () => {});
```
delatrie: Should it be implemented in this PR or in a subsequent one?

packages/service/vitest.config.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export default defineConfig({
+  test: {
+    include: ["./test/**/*.test.ts"],
+    setupFiles: [require.resolve("allure-vitest/setup")],
+    reporters: [
+      "default",
+      [
+        "allure-vitest/reporter",
+        { resultsDir: "./out/allure-results", globalLabels: [{ name: "module", value: "plugin-awesome" }] },
```
delatrie: Typo?
```suggestion
        { resultsDir: "./out/allure-results", globalLabels: [{ name: "module", value: "service" }] },
```

packages/core/src/report.ts
```diff
@@ -189,8 +223,25 @@ export class AllureReport {
...
+          Object.entries(pluginFiles).map(([key, filepath]) =>
+            this.#allureService?.addReportFile({
+              reportUuid: this.#reportUuid,
+              key,
+              filepath,
+            }),
+          ),
+        );
+
+        console.info(`The report has been published: ${this.#reportUrl}`);
```
delatrie: What if we lost the URL (accidentally cleared the console)? Perhaps include the most recent report URL in the output of `allure projects` in case it has been published?

delatrie: Another issue is that the line itself is easy to miss. For example, the `log` plugin configured in `sandbox` emits its output after the URL is printed. Imagine having lots of tests - the line goes up pretty quickly.
My suggestion is to remember we've published the report and print the line at the end so it's guaranteed to be the last one. WDYT?

packages/cli/src/commands/index.ts
```diff
@@ -15,3 +15,7 @@ export * from "./qualityGate.js";
 export * from "./log.js";
 export * from "./generate.js";
 export * from "./dashboard.js";
+export * from "./login.js";
+export * from "./logout.js";
+export * from "./whoami.js";
+export * from "./projects/index.js";
```
delatrie, line 21: All new commands don't support custom configuration files in contrast to, say, `allure generate -c <path>`. That will be inconvenient for people who use multiple configurations (e.g., one for debugging with no history and another with history enabled). I suggest using the same logic as `allure generate` to resolve the config.

packages/service/src/history.ts
```diff
@@ -0,0 +1,19 @@
+import type { AllureHistory, HistoryDataPoint } from "@allurereport/core-api";
+import type { AllureService } from "./service.js";
+
+export class AllureRemoteHistory implements AllureHistory {
+  constructor(readonly allureService: AllureService) {}
+
+  async readHistory(branch?: string) {
+    return await this.allureService.downloadHistory({
```
delatrie: This throws if no project exists, crashing the process. Perhaps, it's okay to return an empty array in such a case. Especially given that the project will be created automatically during publication (if I understand correctly).

packages/cli/src/commands/index.ts
delatrie, line 21: Additionally, those commands issue HTTP requests, which may throw. In such cases, a veeeery long error message with all request details is shown (in my case, it doesn't even fit into the console). I would only show the message and, maybe, a stack trace.

packages/cli/src/commands/projects/init.ts
```diff
@@ -0,0 +1,63 @@
+import { getGitRepoName, readConfig } from "@allurereport/core";
+import { AllureService } from "@allurereport/service";
+import { exit } from "node:process";
+import prompts from "prompts";
+import { green, red } from "yoctocolors";
+import { createCommand } from "../../utils/commands.js";
+
+export const ProjectInitCommandAction = async () => {
```
delatrie: Does the `init` command do something besides creating a project? If not, WDYT about moving the logic into the `project-create`? In other words, make `project-create` retrieve the name from three sources: the CLI, the Repository, and the Prompt. Seems reasonable to me.

delatrie (changes requested):
Some issues I found while exploring a published report:

  - [ ] Allure Service doesn't set `Content-Type` for its JSON responses:

    image


  - [ ] History points have blank names. To reproduce:
    1. Set `accessToken` and `publish: true` in the sandbox's `allurerc.mjs`
    2. Run `yarn workspace sandbox run allure generate ../reader/test/resources/xcresultdata/tags.xcresult` two times
    3. Open the last published report and navigate to the history tab of the test.

        image

  - [ ] Also, I noticed we can't use Allure Service with `allure awesome` command. Does it intentionally not support features that require configuration? If not, I suggest creating an issue to address that in the future. For example, we can change its logic to take the config and disable all plugins except `awesome` (inserting it if it isn't present).

Also, please, see my other comments above.

packages/cli/src/commands/projects/init.ts
epszaw: Yes, it's pretty reasonable

packages/service/README.md
epszaw: I think, we can do this later because at this moment we use the package internally

packages/service/src/utils/http.ts
epszaw: Done

packages/service/src/history.ts
epszaw: Agree

packages/plugin-api/src/config.ts
epszaw: You can place here env variable instead of hardcoded value:

```js
import { env } from "node:process";

export default defineConfig({
  // ...
  allureService: {
	accessToken: env.MY_ACCESS_TOKEN,
  }
})
```

So I think it's better to keep it as simplest as it possible

packages/plugin-awesome/src/generators.ts
epszaw: Yes, create issue, let's fix it outside the PR

epszaw: Same here, let's add issue and solve the problem later

packages/cli/src/commands/index.ts
epszaw: Good catch. Will add typical errors processing to the http-dependent commands

epszaw: Make sense :+1:

packages/plugin-api/src/config.ts
delatrie: Make sense

packages/cli/src/commands/index.ts
epszaw: Done

packages/cli/src/commands/projects/init.ts
epszaw: Done

packages/cli/src/commands/index.ts
epszaw: Done

packages/service/test/index.test.ts
epszaw: Covered the entire package

packages/core/src/report.ts
epszaw: We can find the url on the history tab at any moment

packages/core/src/report.ts
epszaw: :+1:

packages/core/src/report.ts
epszaw: No problem 👌

packages/core/src/report.ts
epszaw: No problem

packages/core/src/history.ts
epszaw: Renamed to the default `acc, item`

epszaw:
> History points have blank names. To reproduce:

XC results reader returns test result without `duration` or `stop` so we can't render anything there. In that case, we can't rely on anything, so I propose to render some kind of placeholder, like `Unknown`

image

packages/core/src/report.ts
delatrie: Doesn't the history tab only contain links to previous reports, but not to the most recent one?

delatrie: As we've discussed, we'll address it in the future.

packages/cli/src/commands/projects/list.ts
```diff
@@ -0,0 +1,107 @@
...
+  } catch (error) {
+    if (error instanceof KnownError) {
+      // eslint-disable-next-line no-console
+      console.error(red(`Failed to get projects: ${error.message}`));
+      process.exit(1);
+      return;
+    }
+
+    if (error instanceof UnknownError) {
+      const logFilePath = await logError("Failed to get projects due to unexpected error", error?.stack);
```
delatrie: Here and in all other commands that use the same error handling strategy:

  - We should log the message of the original error if present (e.g., `logError(error.message ?? "<placeholder>", ...);`
  - No `?.` needed.
  - In some cases, no helpful information is logged at all (e.g., DNS failed, the service is down, or can't connect to the host). The simplest thing we can do is to log `error.code`; it will at least point to the root cause. Another option is to log the entire original error (access token masking required).

packages/service/src/utils/http.ts
```diff
@@ -0,0 +1,89 @@
...
+    super(message);
+    this.name = "UnknownError";
+  }
+}
+
+export const createServiceHttpClient = (
+  historyServiceURL: string = DEFAULT_HISTORY_SERVICE_URL,
+  accessToken?: string,
+) => {
+  const client = axios.create({
```
delatrie, line 36: Should we specify timeout(s)?

packages/service/src/service.ts
```diff
@@ -0,0 +1,200 @@
+import { type HistoryDataPoint } from "@allurereport/core-api";
+import { type Config } from "@allurereport/plugin-api";
+import { readFile } from "node:fs/promises";
+import { join as joinPosix } from "node:path/posix";
+import open from "open";
+import { DEFAULT_HISTORY_SERVICE_URL } from "./model.js";
+import { type HttpClient, createServiceHttpClient } from "./utils/http.js";
+import { decryptExchangeToken, deleteAccessToken, writeAccessToken, writeExchangeToken } from "./utils/token.js";
+
+export class AllureService {
```
delatrie: This is not a major issue, but each time I read the name of this class, I have a thought in my head: "Oh, you really should read it as `AllureSerivceClient`. WDYT about renaming the class to better reflect its purpose?

packages/cli/src/commands/projects/list.ts
```diff
@@ -0,0 +1,107 @@
...
+type CommandOptions = {
+  config?: string;
+  cwd?: string;
+};
+
+export const ProjectsListCommandAction = async (options?: CommandOptions) => {
+  const { config: configPath, cwd } = options ?? {};
+  const config = await readConfig(cwd, configPath);
+
+  if (!config?.allureService?.url) {
```
delatrie, line 17: We should also check if an access token is configured. If not, we should suggest creating it via `allure login`.

We can do this now or in a follow-up PR, though.

packages/service/src/service.ts
```diff
@@ -0,0 +1,200 @@
...
+  async projects() {
+    return this.#client.get<{ id: string; name: string }[]>("/api/projects/list");
+  }
+
+  /**
+   * Creates a new project
+   * @param payload
+   */
+  async createProject(payload: { name: string }) {
+    return this.#client.post<{ id: string; name: string }>("/api/projects/create", {
```
delatrie, line 98: We should catch errors thrown by the server for well-known violations and instead throw domain-specific errors. This will improve the user experience.

For example, for `createProject`, those domain-specific errors are `Project 'foo' already exists` or something like `Bad project name`.

Otherwise, users will be confused by messages like `Failed to create project: Request failed with status code 400`.

We may extract this work in a separate issue.

delatrie (changes requested):
The error handling needs improvement. While it's mostly a future work, there are a couple of issues I would address in this PR. Please, check my comments.

Once we're done with this PR I will create a bunch of issues we've mentioned here and there in this PR.

packages/service/src/service.ts
epszaw: I don't have any concern to rename the class

epszaw: Done

packages/service/src/utils/http.ts
epszaw: I think, it's enough to have the default one here. At least for now

packages/cli/src/commands/projects/list.ts
epszaw: We have two valid options of the token definition:

- the token is provided via config in the `allureService.accessToken` field (the option you're talking about)
- the token exists in the `~/.allure` directory – user logged in and use a personal one

And we can't validate the token's existence here because there are multiple sources of the token.
Additionally, we can't implement the validation logic because not every endpoint requires authentication. In case of strict checking we won't be able to let user to authenticate.

packages/service/src/utils/http.ts
delatrie: Agree. Especially given we don't know yet how much time most requests will require. But we should keep in mind that axios doesn't apply any timeouts by default.

packages/cli/src/commands/projects/list.ts
delatrie: Ok, got it, thanks for the clarification. We can then display such a hint in the case of an HTTP 401 error. That would be good for feature discoverability in cases where a user clones a repo but hasn't signed up to the Allure Service yet.

Anyway, that's out of the scope of this PR. More of a note to myself to create an issue.

packages/cli/src/commands/projects/list.ts
epszaw: - If an error is not a known one – we just throw the error as is, without any modification, so user is able to see, whats going wrong.
- In case of http errors where status < 500 – it's a known error, we can' print the original message from the server because we defined it. I also added status to the console message.
- In case of http errors where status >= 500 – it's an unknown error, we log the original error message (I've already added the logic) with stack and write error's message with the response code.
Sounds ok?

packages/service/src/service.ts
epszaw: It should work as you described. We're responding with a message and status and then, if the status !== 500 we render the message as is.

packages/cli/src/commands/projects/list.ts
delatrie: We shouldn't allow unhandled exceptions to be thrown from the commands, as it creates a potential side channel.

Additionally, the messages from the server are neither displayed to users nor logged in the file. They need to be properly extracted from the response attached to the error.

Other than that, the scheme you've described looks good.

packages/service/src/service.ts
delatrie: I see now. The messages already exist. They just need to be properly extracted on the Allure 3 side. Since we have another conversation about that, I will resolve this one.

delatrie (approved):

PR #154, opened 2025-04-18
feat/#29 add retries usage

Viperwow:


packages/web-components/src/components/Tree/TreeItemInfo/styles.scss
```diff
@@ -0,0 +1,14 @@
+.item-info {
+    display: flex;
```
epszaw, line 2: Run prettier for the project files

packages/web-components/src/components/Tree/TreeItemRetries/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { SvgIcon, allureIcons } from "@/components/SvgIcon";
+import { Text } from "@/components/Typography";
+import styles from "./styles.scss";
+
+export interface TreeItemRetriesProps {
+  retriesCount?: number;
+}
+
+export const TreeItemRetries = ({ retriesCount }: TreeItemRetriesProps) => {
+  if (retriesCount && retriesCount > 0) {
```
epszaw: Let's just set the default value `0` for the property

packages/core/src/utils/stats.ts
```diff
@@ -1,22 +1,26 @@
...
-      if (filter && !filter(test)) {
+  return trsToProcess.reduce<Statistic>(
+    (acc, tr) => {
+      if (filter && !filter(tr)) {
         return acc;
       }
 
-      if (!acc[test.status]) {
-        acc[test.status] = 0;
+      if (tr.retries && tr.retries?.length > 0) {
```
epszaw, line 12: Just `if (tr.retries?.length > 0) {` is enough

epszaw (approved):

packages/web-components/src/components/Tree/TreeItemRetries/index.tsx
Viperwow: I decided to show the retries count if it is provided

packages/core/src/utils/stats.ts
Viperwow: You cannot compare possible `undefined` with numeric value except equation

packages/web-components/src/components/Tree/TreeItemRetries/index.tsx
Viperwow: An, no, it's okay. Existence check is on another level :)

PR #164, opened 2025-04-30
feat/#157-add-flaky-indication

Viperwow:
- Add tree-based flaky indicator
- Add flaky indicator usage in Summary metadata
- Add tests for flaky test results
- Fix "testSummary.retries" locale value
- Split tests accordingly to feature ("flaky", "retries")
- Update flaky icon to be the same as existing icons


packages/core/src/utils/flaky.ts
```diff
@@ -0,0 +1,18 @@
...
+const isAllureClassicFlaky = (tr: TestResult, history: HistoryTestResult[]) => {
+    if (history.length === 0 || !badStatuses.includes(tr.status)) {
+        return false;
+    }
+
+    const limitedLastHistory = history.slice(-MAX_LAST_HISTORY_SIZE);
+    const limitedLastHistoryStatuses = limitedLastHistory.map((h) => h.status);
+
+    return limitedLastHistoryStatuses.includes("passed")
+    && limitedLastHistoryStatuses.indexOf("passed") < limitedLastHistoryStatuses.lastIndexOf("failed");
```
delatrie, line 15: Should we also account for `broken` here? A test that goes `broken` -> `passed` -> `broken` looks flaky to me.

packages/e2e/test/allure-awesome/tree/flaky.test.ts
```diff
@@ -0,0 +1,247 @@
...
+        },
+        {
+          name: flakyTestName,
+          fullName: flakyTestFullname,
+          historyId,
+          status: Status.PASSED,
+          stage: Stage.FINISHED,
+          start: now + 4000,
+          stop: now + 5000,
+        },
```
delatrie: You've already defined the `failed` -> `failed` -> `passed` history for this test. No need to define retries in addition to that.

packages/core/src/utils/flaky.ts
Viperwow: This is the same implementation as in https://github.com/allure-framework/allure2/blob/ecb18585042431f9806c667bfda3d86b3fccd185/allure-generator/src/main/java/io/qameta/allure/history/HistoryPlugin.java#L113

But I can agree that adding `broken` is the right decision

Viperwow: We decided to stick with Allure2 implementation and make changes later on.

delatrie (approved):

PR #168, opened 2025-05-05
Introduce page objects

epszaw:


packages/e2e/test/allure-awesome/pageObjects/Common.ts
```diff
@@ -0,0 +1,32 @@
...
+  reportTitleLocator: Locator;
+
+  toggleLayoutButtonLocator: Locator;
+  splitLayoutLocator: Locator;
+  singleLayoutLocator: Locator;
+
+  envPickerLocator: Locator;
+  envPickerButtonLocator: Locator;
+
+  constructor(page: Page) {
```
Viperwow: missing readonly?

packages/e2e/test/allure-awesome/test/tree.test.ts
Viperwow, line 1: It looks like it might be missing the check for retries marker & count in the "retries" section. Or missing the entire file tests

packages/e2e/test/allure-awesome/pageObjects/Tree.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  async searchTree(text: string) {
+    await this.searchLocator.fill(text);
+  }
+
+  async searchClear() {
+    await this.searchClearLocator.click();
+  }
+
+  async clickRetryFilter() {
```
Viperwow: Can't we split the opening Filter Modal out into a separate method? It can be helpful in possible tests related to Filter

packages/e2e/test/allure-awesome/test/tree.test.ts
```diff
@@ -0,0 +1,614 @@
...
+    await expect(treePage.leafLocator).toHaveCount(3);
+  });
+
+  test("should show retry icon in the tree for tests with retries", async ({ page }) => {
+    const retryIcons = page.getByTestId("tree-item-retries");
+
+    await expect(retryIcons).toHaveCount(2);
+
+    const testWithRetriesIcon = treePage.leafLocator
+      .filter({
```
Viperwow: There might be a method for selecting the tree item with the specific text. It is being used widely in flaky, retries, new, etc. tests.

 What do you think?

Viperwow (changes requested):
Add missing **flaky** tests

packages/e2e/test/allure-awesome/test/tree.test.ts
epszaw: We already have the method. Will use it

PR #170, opened 2025-05-06
Web-summary updates

todti:
- new summary card view
- add Info function in some plugins to make card more detailed 
- some style improvements


packages/core/src/report.ts
```diff
@@ -199,7 +199,7 @@ export class AllureReport {
 
       summaries.push({
         ...summary,
-        href: join("/", id),
+        href: join("/", id, "/"),
```
epszaw: `join` can produce a wrong path on windows. Use `node:path/posix` or `URL` class to build the link

epszaw (changes requested):

PR #180, opened 2025-05-15
Fix visually truncated error message in test result for awesome plugin

evg-fr:
On test result page error message the text is not wrapped resulting in visual truncation (actual text is there, but hidden).
Seems that default styling for `<pre>` tags doesn't respect text wrapping.

Without fix:
allure3_bug

With fix:
allure3_fixed


todti:
Hi @evg-fr 
Thank you for your contribute but white-space breaks error message and difficult to read error trace here. We use white-space by default for this error messages. If you need some test results you can use `sandbox` directory just run `yarn t` inside and you get some `allure-results` and `alllure-report`

image
image

evg-fr:
Hello @todti,
It's interesting because in results that we got with NUnit Allure 2.9.5-preview (it's a bit old, but we stuck with it for some time) package we have separate field for stack trace and it shows it with scrollbar preserving line-breaks:
image
image

test result json has separate field for stack trace:
image

I'm a bit confused because there was fix for it: #32

denysgaievskyi:
hey, 

from what i can see, the problem is still with us and is very annoying. any plans to merge this fix? @todti

todti (approved):

PR #196, opened 2025-05-30
Feat/#159 add new tests indication

Viperwow:


packages/core/src/utils/new.ts
```diff
@@ -0,0 +1,51 @@
...
+export const isRegressed = (tr: TestResult, history: HistoryTestResult[] = []) =>
+  isHistoricallyNew(tr, history) && tr.status === "failed";
+
+export const isMalfunctioned = (tr: TestResult, history: HistoryTestResult[] = []) =>
+  isHistoricallyNew(tr, history) && tr.status === "broken";
+
+export const getStatusTransition = (
+  tr: TestResult,
+  history: HistoryTestResult[] = [],
+): TestStatusTransition | undefined => {
```
epszaw, line 39: Don't we want to turn `TestStatusTransition` into `enum`?

epszaw (approved):

packages/core/src/utils/new.ts
Viperwow: Let's stay with this for consistency among TestStatus / TestStatusTransition

packages/core/src/utils/new.ts
```diff
@@ -0,0 +1,51 @@
...
+const getMeaningfulHistory = (history: HistoryTestResult[] = []): HistoryTestResult[] =>
+  history.filter((htr) => !NON_MEANINGFUL_HISTORY_STATUSES.includes(htr.status));
+
+/**
+ * @description Checks if the test result is switched to another meaningful status.
+ * @param tr - The test result to check.
+ * @param history - The history of test results.
+ * @returns `true` if the test result is switched to a new status, `false` otherwise.
+ */
+export const isHistoricallyNew = (tr: TestResult, history: HistoryTestResult[] = []) => {
```
delatrie: IMO, something like `hasNewStatus`, or `hasStatusChange`, or `hasStatusTransition` works better. I wouldn't guess the difference between `isNew` and `isHistoricallyNew` based solely on the function names.

packages/core/src/utils/new.ts
```diff
@@ -0,0 +1,51 @@
...
+  const meaningfulHistory = getMeaningfulHistory(history);
+
+  // If status is only current on the test result, it is new.
+  if (meaningfulHistory.length === 0) {
+    return true;
+  }
+
+  return meaningfulHistory[meaningfulHistory.length - 1].status !== tr.status;
```
delatrie: Please note that the history array is sorted from most recent to least recent entries. That means `meaningfulHistory[meaningfulHistory.length - 1]` gives the earliest entry, not the most recent one.

packages/core/src/utils/new.ts
```diff
@@ -0,0 +1,51 @@
...
+export const isFixed = (tr: TestResult, history: HistoryTestResult[] = []) =>
+  isHistoricallyNew(tr, history) && tr.status === "passed";
+
+export const isRegressed = (tr: TestResult, history: HistoryTestResult[] = []) =>
+  isHistoricallyNew(tr, history) && tr.status === "failed";
+
+export const isMalfunctioned = (tr: TestResult, history: HistoryTestResult[] = []) =>
+  isHistoricallyNew(tr, history) && tr.status === "broken";
+
+export const getStatusTransition = (
```
delatrie, line 36: The implementation copies the entire history of the test result 4 times. That results in a complexity of `N * D`, where `N` is the number of test results and `D` is the depth of the history. Since history isn't limited in Allure 3, such complexity may become quite significant.

I suggest using a slightly less complex algorithm:

1. Retrieve the most recent `passed/failed/broken` entry (no need to copy the entire array; iterate from 0 until we find a suitable entry).
2. If it's undefined -> `new`
3. If its status doesn't equal the current status, assign a tag based on the current status (passed -> `fixed`, failed -> `regressed`, broken -> `malfunctioned`).

The result is the same, but we'll avoid iterating through the entire history for the vast majority of cases. While the worst-case scenario still gives us the `N * D` complexity, the average time will be an order of magnitude better.

packages/core/src/utils/new.ts
```diff
@@ -0,0 +1,51 @@
+import type { HistoryTestResult, TestResult, TestStatusTransition } from "@allurereport/core-api";
+
+const NON_MEANINGFUL_HISTORY_STATUSES = ["unknown", "skipped"];
```
delatrie: WDYT about using "significant" instead of "meaningful"? I mean, the `skipped` status is not meaningless. It has the meaning (the test was skipped). But it's not significant (at least in the context of status transitions).

packages/e2e/test/allure-awesome/tree/transitions.test.ts
```diff
@@ -0,0 +1,248 @@
...
+            uuid: getRandomUUID(),
+            name: reportName,
+            timestamp: 1746612096098,
+            knownTestCaseIds: [newFailedTestTestCaseId, newPassedTestTestCaseId, newBrokenTestTestCaseId],
+            testResults: {
+              [newFailedTestHistoryId]: {
+                id: getRandomUUID(),
+                name: newFailedTestName,
+                fullName: newFailedTestFullname,
+                status: Status.PASSED,
```
delatrie: Let's change the statuses of the earliest history entries to match the current statuses. That way, the tests will cover the bug in `new.ts` (I mentioned it in one of my previous comments): https://github.com/allure-framework/allure3/blob/93dea5db97e0251309eaf5176d8fe7e02709773a/packages/core/src/utils/new.ts#L24

We may also set it to `skipped/unknown` for one test to cover the handling of such statuses.

```suggestion
                status: Status.FAILED,
```

packages/web-awesome/src/components/ReportBody/Filters.tsx
```diff
@@ -1,12 +1,48 @@
...
   const { t } = useI18n("filters");
-  const { flaky, retry, new: isNew } = treeFiltersStore.value.filter;
-  const hasFilter = flaky || retry || isNew;
+  const hasFilter = MENU_KEYS.some((key) => treeFiltersStore.value.filter[key]);
+
+  const renderFilterItem = (filter: TreeFilters, value: boolean) => {
+    return (
+      <Menu.Item
+        closeMenuOnClick={false}
+        ariaLabel={t("enable-filter", { filter: t(filter) })}
```
delatrie, line 25: Is it possible to also add a tooltip that explains what tests the filter keeps?

packages/web-components/src/components/Tree/TreeItemInfo/index.tsx
```diff
@@ -1,24 +1,38 @@
...
   const formattedDuration = formatDuration(duration);
 
   return (
     <div className={styles["item-info"]}>
-      {newTest && <TreeItemMetaIcon type="new" />}
-      {flakyTest && <TreeItemMetaIcon type="flaky" />}
+      {flaky && <SvgIcon data-testid="tree-leaf-flaky" id={allureIcons.lineIconBomb2} />}
       <TreeItemRetries retriesCount={retriesCount} />
+      {transition && (
+        <Tag data-testid={`tree-leaf-transition-${transition}`} skin={transitionToTagSkin(transition)}>
```
delatrie, line 32: Is it possible to add a tooltip to tags? I can imagine people wondering what exactly `regressed` or `malfunctioned` means.

delatrie (changes requested):

packages/web-awesome/src/components/ReportBody/Filters.tsx
Viperwow: Same as in Tags

packages/e2e/test/allure-awesome/tree/transitions.test.ts
Viperwow: Fixed this in a implementation and history is presorted on a Plugin side, there is no need to change tests order

packages/web-components/src/components/Tree/TreeItemInfo/index.tsx
Viperwow: Will be implemented in another task
https://github.com/allure-framework/allure-org/issues/85

packages/web-awesome/src/components/ReportBody/Filters.tsx
Viperwow: https://github.com/allure-framework/allure-org/issues/85

packages/core/src/utils/new.ts
Viperwow: Done

packages/core/src/utils/new.ts
Viperwow: Done

packages/e2e/test/allure-awesome/tree/transitions.test.ts
delatrie: My intention was to cover the following behavior:

  - The most recent history entry is checked first
  - History entries with insignificant statuses are ignored

The current e2e test doesn't check either. I see two options to fix the coverage:

  - Tweak the e2e test slightly: change the earliest statuses of some tests to match their current statuses, and set the oldest statuses of some other tests to `skipped`. This is what I've originally proposed.
  - Keep the e2e test as is and cover the exported functions of `new.ts` with unit tests. This may be a better alternative as it doesn't complicate the e2e tests, and it's also much faster.

Viperwow: I've introduced some data that will cover Skipped status and history order & history sample length (5+)

delatrie (approved):

PR #206, opened 2025-06-05
fix: round percentage values to two decimal places in charts calculations

Viperwow:


packages/plugin-awesome/src/charts.ts
```diff
@@ -245,7 +245,7 @@ const calculatePercentValues = <T extends TrendDataType>(
 
     points[pointId] = {
       x: executionId,
-      y: (value / total) * 100,
+      y: Number(((value / total) * 100).toFixed(2)),
```
delatrie, line 248: Is it possible to leave the original values intact while formatting them for display purposes? Looks like that's supported via the legend's [`data` property](https://github.com/plouc/nivo/blob/a2d9dab855365926cb41267eb20af154ca8fd558/packages/legends/src/types.ts#L69).

If so, I would format the values up to two significant digits instead of some fixed precision:

```js
const valueToDisplayPercentage = value.toLocaleString(undefined, { style: 'percent', maximumSignificantDigits: 2 }); // with '%'
const valueToDisplay = value.toLocaleString(undefined, { maximumSignificantDigits: 2 }); //without '%'
```

Otherwise, it's still preferable to round up to significant digits IMO:

```js
const y = parseFloat((value * 100 / total).toLocaleString(undefined, { maximumSignificantDigits: 2 }));
```

epszaw (approved):

delatrie (changes requested):

PR #212, opened 2025-06-11
some small fixes

todti:
just notice some bugs

### DIFF
--- a/packages/web-components/src/components/Attachment/styles.scss
+++ b/packages/web-components/src/components/Attachment/styles.scss
@@ -24,6 +24,7 @@
 
 .test-result-attachment-content {
   border-top: 1px solid var(--on-border-muted);
+  cursor: pointer;
 }
 
 .test-result-attachment-content-wrapper {
@@ -150,7 +151,6 @@
 }
 
 .test-result-attachment-image {
-  cursor: pointer;
   padding: 4px 0;
   position: relative;
   width: 100%;


packages/web-awesome/src/components/MetadataButton/index.tsx
```diff
@@ -27,7 +27,7 @@ export const MetadataButton: FunctionalComponent<MetadataButtonProps> = ({
       <Text size={"m"} bold>
         {title}
       </Text>
-      {counter !== null && <Counter count={counter} size="s" />}
+      {counter !== undefined && <Counter count={counter} size="s" />}
```
epszaw: Why can't we use just `!!counter`? It'll handle both negative cases

packages/web-awesome/src/components/TestResult/TrSteps/TrAttachment.tsx
```diff
@@ -96,7 +108,7 @@ export const TrAttachment: FunctionComponent<{
       </div>
       {isOpened && isValidComponentType && (
         <div className={styles["test-result-attachment-content-wrapper"]}>
-          <div className={styles["test-result-attachment-content"]}>
+          <div className={styles["test-result-attachment-content"]} onClick={expandAttachment}>
```
epszaw: Add `role` to the element. I don't like the idea of using `div` elements as interactive ones because it's a such of antipattern and absolutely inaccessible thing. Can't we use a button here? Or for example, wrap into `button` tag just title?

epszaw (approved):

PR #213, opened 2025-06-13
Feat/#85-add-tooltips-for-transitions

Viperwow:
#85


packages/web-awesome/src/locales/en.json
```diff
@@ -33,6 +33,14 @@
     "regressed": "Regressed",
     "malfunctioned": "Malfuctioned"
   },
+  "filters.description": {
+    "flaky": "Show test results that \"passed\" and then \"failed\" within the last 5 runs, indicating unstable behavior",
```
delatrie: I wouldn't go into the details of the specific algorithm, given that we will allow the use of custom ones. Usually, people know what "flaky" means because it's a well-established term in the industry, unlike Allure-specific terms (such as "regressed" and "malfunctioned"). So, we can use a shorter description, such as "Show unstable (flaky) tests."

packages/web-awesome/src/locales/en.json
```diff
@@ -33,6 +33,14 @@
...
+    "retry": "Show test results that were rerun with no status changes",
+    "new": "Show test results that appear for the first time in this report",
+    "fixed": "Show test results that changed to \"passed\" status from \"failed\" or \"broken\" status",
+    "regressed": "Show test results that changed to \"failed\" status from \"passed\" status",
```
delatrie: According to the current implementation, not only from "passed", but also from "broken"

packages/web-awesome/src/locales/en.json
```diff
@@ -164,5 +172,11 @@
     "fixed": "Fixed",
     "regressed": "Regressed",
     "malfunctioned": "Malfuctioned"
+  },
+  "transitions.description": {
+    "new": "First occurrence of this test result in the report",
```
delatrie: nit:
```suggestion
    "new": "The first occurrence of this test result in the report",
```

packages/web-awesome/src/locales/en.json
```diff
@@ -164,5 +172,11 @@
...
+    "fixed": "Previously \"failed\" or \"broken\" test that is now \"passed\"",
```
delatrie: nit:
```suggestion
    "fixed": "A previously \"failed\" or \"broken\" test that is now \"passed\"",
```

packages/web-awesome/src/locales/en.json
```diff
@@ -164,5 +172,11 @@
...
+    "regressed": "Previously \"passed\" test that is now \"failed\"",
```
delatrie: nit:
```suggestion
    "regressed": "A previously \"passed\" or \"broken\" test that is now \"failed\"",
```

packages/web-awesome/src/locales/en.json
```diff
@@ -164,5 +172,11 @@
...
+    "malfunctioned": "Previously \"passed\" test that is now \"broken\""
```
delatrie: bit:
```suggestion
    "malfunctioned": "A previously \"passed\" or \"failed\" test that is now \"broken\""
```

packages/web-awesome/src/locales/en.json
```diff
@@ -33,6 +33,14 @@
     "regressed": "Regressed",
     "malfunctioned": "Malfuctioned"
   },
+  "filters.description": {
+    "flaky": "Show test results that \"passed\" and then \"failed\" within the last 5 runs, indicating unstable behavior",
+    "retry": "Show test results that were rerun with no status changes",
```
delatrie: Currently, the filter shows all retried tests regardless of status changes.

packages/web-awesome/src/locales/en.json
```diff
@@ -33,6 +33,14 @@
...
+    "new": "Show test results that appear for the first time in this report",
+    "fixed": "Show test results that changed to \"passed\" status from \"failed\" or \"broken\" status",
+    "regressed": "Show test results that changed to \"failed\" status from \"passed\" status",
+    "malfunctioned": "Show test results that changed to \"broken\" status from \"passed\" status"
```
delatrie: According to the current implementation, not only from "passed", but also from "failed"

packages/web-awesome/src/locales/en.json
```diff
@@ -33,6 +33,14 @@
     "regressed": "Regressed",
     "malfunctioned": "Malfuctioned"
   },
+  "filters.description": {
+    "flaky": "Show test results that \"passed\" and then \"failed\" within the last 5 runs, indicating unstable behavior",
+    "retry": "Show test results that were rerun with no status changes",
+    "new": "Show test results that appear for the first time in this report",
+    "fixed": "Show test results that changed to \"passed\" status from \"failed\" or \"broken\" status",
```
delatrie: This wording is OK, but it would be better if we could clarify that we're comparing the status in the current report with the status in an earlier report. Otherwise, people may get confused by thinking we're dealing with retries here.

For example:

`Show tests that are passed now but have been failed or broken in a previous report.`

delatrie (changes requested):

delatrie (approved):

PR #217, opened 2025-06-17
Add ability to filter log plugin entries

epszaw:


packages/plugin-log/test/utils.test.ts
```diff
@@ -1,109 +1,292 @@
...
+    });
   });
-});
 
-describe("stringifyStatusBadge", () => {
-  it("returns green badge for passed status", () => {
-    expect(stringifyStatusBadge("passed")).toBe("\u001b[32m✓\u001b[39m");
-  });
+  describe("stringifyStatusBadge", () => {
+    it("returns green badge for passed status", () => {
```
Viperwow, line 70: Isn't it better to write a parameterised test, but I do not insist?

epszaw: It makes sense, but we can do it later, this implementation it's a previous, unchanged one

PR #223, opened 2025-06-24
ansi support for attachments

### LINKED ISSUE #218: ANSI color escape codes not recognized in logs
**Describe the bug**

In log attachments, ANSI color escape codes are not processed: colors are not applied, and escape sequences are displayed as text.

**To Reproduce**

1. Have a pytest test suite with logging
2. Generate results using `allure-pytest==2.14.3`
3. Generate report using allure `3.0.0-beta.15`
4. Open "log" attachment of any test

**Expected behavior**

Logs are rendered with colors, like pytest-html does.

**Screenshots**

Allure:

Allure

pytest-html:

pytest-html

 - OS: Linux
 - Browser: Firefox
 - Version: 139.0.4

**Additional context**

Doesn't work with allure v2 (allure-commandline 2.34.0) too: 

Image

todti:
- closes #218


packages/web-components/src/components/Attachment/Attachment.tsx
```diff
@@ -18,7 +18,7 @@ const componentsByAttachmentType: Record<string, any> = {
   uri: AttachmentCode,
   css: AttachmentCode,
   table: AttachmentCode,
-  html: AttachmentCode,
+  html: HtmlPreview,
```
delatrie: Pls, revert. HTML attachments should be rendered as code (i.e., their markup should be displayed). Embedding them directly into the report is incorrect and won't work in most cases.

todti: thanks for notice! it was unnecessary code to commit

packages/web-awesome/src/components/TestResult/TrSteps/TrAttachment.tsx
```diff
@@ -109,7 +109,7 @@ export const TrAttachment: FunctionComponent<{
       {isOpened && isValidComponentType && (
         <div className={styles["test-result-attachment-content-wrapper"]}>
           <div className={styles["test-result-attachment-content"]} role={"button"} onClick={expandAttachment}>
-            <Attachment item={item} />
+            <Attachment item={item} previewable={true} />
```
delatrie: Currently, this option only affects HTML attachments. It embeds the HTML content into the preview area of the attachment step (not in an expanded dialog) instead of displaying its markup.

However, it's better to leave the markup instead of the preview here since not all HTML attachments can be displayed that way (see #226). Also, we have unsolved problems with XSS (#227). Please, revert this change until we solve those issues.

delatrie (changes requested):
Something inconsistent happens. The transformation only works every other time:

https://github.com/user-attachments/assets/b5ec131e-c4b2-476a-8ae4-a90b0b81e492

delatrie (approved):

epszaw (approved):

PR #240, opened 2025-07-01
Titlepath tree

todti:
- add tree hierarchy by Title path for Allure Awesome plugin as default `groupBy`


packages/reader/src/allure2/index.ts
```diff
@@ -144,8 +144,10 @@ export const allure2: ResultsReader = {
 };
 
 const processTestResult = async (visitor: ResultsVisitor, result: Partial<TestResult>, originalFileName: string) => {
+  console.log("RESULT", result.titlePath);
```
delatrie: Debug leftover?

packages/sandbox/package.json
```diff
@@ -20,8 +20,8 @@
     "@typescript-eslint/parser": "^8.0.0",
     "@vitest/runner": "^2.1.9",
     "allure": "workspace:*",
-    "allure-js-commons": "^3.0.9",
-    "allure-vitest": "^3.0.9",
+    "allure-js-commons": "^3.3.0",
+    "allure-vitest": "^3.3.0",
```
delatrie, line 25: Let's also bump in other packages

packages/core-api/src/model.ts
```diff
@@ -90,6 +92,7 @@ export interface TestFixtureResult {
   steps: TestStepResult[];
 
   sourceMetadata: SourceMetadata;
+  titlePath?: string[];
```
delatrie: What's the motivation for having `titlePath` for fixtures? We don't navigate between fixtures, and we only show them as part of the test results.

packages/reader/src/allure2/model.ts
```diff
@@ -67,6 +67,7 @@ interface ExecutableItem {
   parameters?: Parameter[];
   start?: number;
   stop?: number;
+  titlePath?: string[];
```
delatrie: `titlePath` is test-specific, let's move it into `TestResult`.

packages/plugin-api/src/utils/tree.ts
```diff
@@ -329,3 +329,87 @@ export const transformTree = <L, G>(
...
+
+/**
+ * Creates tree by Title path
+ * @param tests
+ * @param getTitlePath
+ * @param getLeaf
+ */
+export const createTreeByTitlePath = <T extends Partial<TestResult>>(
+  tests: T[],
+  getTitlePath: (item: T) => string[],
```
delatrie: Isn't it always `item.titlePath ?? []`?

packages/plugin-api/src/utils/tree.ts
```diff
@@ -329,3 +329,87 @@ export const transformTree = <L, G>(
...
+  const root: { groups: string[]; leaves: string[] } = { groups: [], leaves: [] };
+  const groupsById: Record<string, any> = {};
+  const leavesById: Record<string, any> = {};
+  const groupPathToId = new Map<string, string>();
+  let groupOrder = 1;
+
+  for (const test of tests) {
+    const path = getTitlePath(test);
+    if (!Array.isArray(path) || path.length === 0) {
+      continue;
```
delatrie: Let's fall back to `[]` instead of dropping the test result altogether.

packages/plugin-api/src/utils/tree.ts
```diff
@@ -329,3 +329,87 @@ export const transformTree = <L, G>(
...
+/**
+ * Creates tree by Title path
+ * @param tests
+ * @param getTitlePath
+ * @param getLeaf
+ */
+export const createTreeByTitlePath = <T extends Partial<TestResult>>(
+  tests: T[],
+  getTitlePath: (item: T) => string[],
+  getLeaf: (item: T) => any & { nodeId: string; status: string },
```
delatrie: What's the motivation for making this function injectable?

packages/plugin-api/src/utils/tree.ts
```diff
@@ -329,3 +329,87 @@ export const transformTree = <L, G>(
...
+    if (!Array.isArray(path) || path.length === 0) {
+      continue;
+    }
+
+    let parentGroupId: string | undefined;
+    const pathParts: string[] = [];
+
+    for (const part of path) {
+      pathParts.push(part);
+      const key = pathParts.join("/");
```
delatrie: Collisions are possible if there is `/` in a title path. For example, the following four title paths:

  - `["foo/bar/baz"]`
  - `["foo/bar", "baz"]`
  - `["foo", "bar/baz"]`
  - `["foo", "bar", "baz"]`

...will all give the `foo/bar/baz` key.

Let's use `JSON.stringify` instead. It's collision-free:

```js
const key = JSON.stringify(pathParts);
```

packages/plugin-api/src/utils/tree.ts
```diff
@@ -53,7 +53,7 @@ const createTree = <T, L, G>(
     let parentGroups = [root];
 
     for (const layer of itemGroups) {
-      if (layer.length === 0) {
+      if (layer?.length === 0) {
```
epszaw: Better to check `!layer?.length` because if the `length` is not available `layer?.length` equals to `undefined` and the condition is falsy

packages/reader/src/allure2/model.ts
epszaw: By the way, you can use `yarn upgrade-interactive` command at the root of the project. It updates the chosen dependency across the whole project

packages/plugin-awesome/README.md
```diff
@@ -47,11 +47,12 @@ export default defineConfig({
...
+| Option           | Description                                     | Type    | Default                 |
+|------------------|-------------------------------------------------|---------|-------------------------|
+| `reportName`     | Name of the report                              | `string` | `Allure Report`         |
+| `singleFile`     | Writes the report as a single `index.html` file | `boolean` | `false`                 |
+| `logo`           | Path to the logo image                          | `string` | `null`                  |
+| `theme`          | Default color theme of the report               | `light \| dark` | OS theme                |
+| `reportLanguage` | Default language of the report                  | `string` | OS language             |
+| `ci`             | CI data which will be rendered in the report    | `{ type: "github" \| "jenkins", url: string, name: string }` | `undefined`             |
+| `groupBy`        | By default, tests are grouped using the `titlePath` provided by the test framework.    | `string`| Grouping by `titlepath` |
```
delatrie, line 58: > By default, tests are grouped using the `titlePath` provided by the test framework.

Let's also explain what this parameter does, not only what the default value is.

packages/reader/src/allure2/index.ts
```diff
@@ -146,6 +146,7 @@ export const allure2: ResultsReader = {
 const processTestResult = async (visitor: ResultsVisitor, result: Partial<TestResult>, originalFileName: string) => {
   const dest: RawTestResult = {
     uuid: ensureString(result.uuid),
+    titlePath: result?.titlePath?.length ? result.titlePath : [],
```
delatrie, line 149: A reminder for me to create an issue for a fallback.

packages/plugin-awesome/src/plugin.ts
```diff
@@ -37,11 +37,16 @@ export class AwesomePlugin implements Plugin {
...
-      !groupBy.length ? ["parentSuite", "suite", "subSuite"] : groupBy,
-      convertedTrs,
-      ({ labels }) => labels.map(({ name }) => name),
-    );
+    const hasGroupBy = groupBy.length > 0;
+
+    const treeLabels = hasGroupBy
+      ? preciseTreeLabels(groupBy, convertedTrs, ({ labels }) => labels.map(({ name }) => name))
+      : [];
+    // const treeLabels = preciseTreeLabels(
```
epszaw: Leftovers?

epszaw (approved):

PR #243, opened 2025-07-03
Add report's remote href to the summary file

epszaw:


packages/core/src/report.ts
```diff
@@ -248,12 +248,26 @@ export class AllureReport {
         return;
       }
 
+      if (context.publish) {
+        summary.remoteHref = `${this.reportUrl}/${context.id}/`;
+        console.info(`The report for "${context.id}" plugin has been published and availabe at: ${summary.remoteHref}`);
```
delatrie: Let's remember the plugin is published, and print the list of published plugins at the end. Otherwise, that information will be hard to find in the console output.

packages/web-awesome/webpack.config.js
```diff
@@ -1,6 +1,7 @@
 import ForkTsCheckerPlugin from "fork-ts-checker-webpack-plugin";
 import HtmlWebpackPlugin from "html-webpack-plugin";
 import MiniCssExtractPlugin from "mini-css-extract-plugin";
+import { randomBytes } from "node:crypto";
```
delatrie: This import can be removed

packages/web-allure2/webpack.config.js
```diff
@@ -16,8 +16,8 @@ export default (env, argv) => {
     entry: "./src/index.js",
     output: {
       path: join(__dirname, SINGLE_FILE_MODE ? "dist/single" : "dist/multi"),
-      filename: devMode ? "app.js" : "app-[fullhash].js",
-      assetModuleFilename: `[name][ext]`,
+      filename: "app-[fullhash].js",
```
delatrie: Why does this line differ from all other web plugins?

delatrie (approved):

packages/web-awesome/webpack.config.js
epszaw: Removed

packages/web-allure2/webpack.config.js
epszaw: Fixed

packages/core/src/report.ts
epszaw: Done

PR #263, opened 2025-07-16
Migrate CLI

epszaw:


packages/cli/src/commands/awesome.ts
```diff
@@ -1,138 +1,104 @@
...
+  });
+
+  resultsDir = Option.String({ required: true, name: "The directory with Allure results" });
+
+  config = Option.String("--config,-c", {
+    description: "The path Allure config file",
+  });
+
+  cwd = Option.String("--cwd", {
+    description: "The working directory for the command to run (Default: current working directory)",
```
Viperwow: Inconsistent casing in "Default" => "default"

packages/cli/src/commands/projects/create.ts
```diff
@@ -1,103 +1,103 @@
...
+      console.info(lines.join("\n"));
+    } catch (error) {
+      if (error instanceof KnownError) {
+        // eslint-disable-next-line no-console
+        console.error(red(error.message));
+        exit(1);
+        return;
+      }
+
+      await logError("Failed to create project due to unexpected error", error as Error);
```
Viperwow, line 99: Some of the plugins have `logError` and `console.error()`. We just log unexpected errors?

Viperwow: Discussed it, now it's okay to use this type of logging as in Wrangler

packages/cli/src/commands/awesome.ts
epszaw: Done

PR #264, opened 2025-07-21
Complete quality gate, add pack/unpack commands for test results

epszaw:
complete quality gate reporting

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -46,7 +46,7 @@ export class GenerateCommand extends Command {
 
   async execute() {
     const cwd = this.cwd ?? process.cwd();
-    const resultsDir = this.resultsDir ?? "./**/allure-results";
+    const resultsDir = (this.resultsDir ?? "./**/allure-results").replace(/[\\/]$/, "");
     const config = await readConfig(cwd, this.config, {
       name: this.reportName,
       output: this.output ?? "allure-report",
--- a/packages/cli/src/commands/qualityGate.ts
+++ b/packages/cli/src/commands/qualityGate.ts
@@ -63,7 +63,7 @@ export class QualityGateCommand extends Command {
 
   async execute() {
     const cwd = await realpath(this.cwd ?? processCwd());
-    const resultsDir = this.resultsDir ?? "./**/allure-results";
+    const resultsDir = (this.resultsDir ?? "./**/allure-results").replace(/[\\/]$/, "");
     const { maxFailures, minTestsCount, successRate, fastFail, knownIssues: knownIssuesPath } = this;
     const config = await readConfig(cwd, this.config, {
       knownIssuesPath,
--- a/packages/cli/src/commands/results/pack.ts
+++ b/packages/cli/src/commands/results/pack.ts
@@ -60,7 +60,7 @@ export class ResultsPackCommand extends Command {
 
   async execute() {
     const cwd = await realpath(this.cwd ?? process.cwd());
-    const resultsDir = this.resultsDir ?? "./**/allure-results";
+    const resultsDir = (this.resultsDir ?? "./**/allure-results").replace(/[\\/]$/, "");
     const archiveName = this.name ?? "allure-results.zip";
     const resultsDirectories = new Set<string>();
     const resultsFiles = new Set<string>();
--- a/packages/cli/src/commands/run.ts
+++ b/packages/cli/src/commands/run.ts
@@ -144,9 +144,9 @@ const runTests = async (params: {
   const afterProcess = Date.now();
 
   if (code !== null) {
-    console.log(`process finished with code ${code} (${afterProcess - beforeProcess})ms`);
+    console.log(`process finished with code ${code} (${afterProcess - beforeProcess}ms)`);
   } else {
-    console.log(`process terminated (${afterProcess - beforeProcess})ms`);
+    console.log(`process terminated (${afterProcess - beforeProcess}ms)`);
   }
 
   await allureResultsWatch.abort();


packages/cli/src/commands/run.ts
```diff
@@ -81,22 +83,32 @@ const runTests = async (
...
+  return {
+    process: testProcess,
+    // FIXME: at this moment, this is the cheapest way to have the process to control and keep ability to receive original exit code
+    // eslint-disable-next-line no-async-promise-executor
+    processTerminationPromise: new Promise(async (res) => {
+      const code = await terminationOf(testProcess);
+      const afterProcess = Date.now();
 
-  await allureResultsWatch.abort();
+      console.log(`process finished with code ${code ?? 0} (${afterProcess - beforeProcess})ms`);
```
delatrie: This message is misleading if the process is terminated due to a signal (e.g., because the fast fail quality gate rule has been triggered). In that case, the message will tell that the return code is 0, which conventionally means success.

In reality, in such a case, the exit code can (and in most cases will) be different (because hitting a signal usually means a failure). Furthermore, Node.js doesn't reveal the actual exit code. It sets `code` to `null` instead. From [here](https://nodejs.org/api/child_process.html#event-exit):

> code <number> The exit code if the child process exited on its own, or null if the child process terminated due to a signal.

I suggest extracting the value of `signal` in addition to `code` from `process.on("exit", (code, signal) => { ... })` and in case `signal` is not `null`, displaying the message `with signal ${signal}` instead of `with code 0`.

packages/cli/src/commands/generate.ts
```diff
@@ -20,7 +21,10 @@ export class GenerateCommand extends Command {
     ],
   });
 
-  resultsDir = Option.String({ required: true, name: "The directory with Allure results" });
+  resultsDir = Option.String({
+    required: false,
+    name: "The directory with Allure results. Entire working directory will be scanned for `allure-results` unless this option is provided.",
```
delatrie: How about making this argument a glob pattern? That way, if the argument is not provided, the default pattern `./**/allure-results/` will be used, so both behaviors (with and without the argument) become consistent with each other.

Currently, the behavior without the argument is very unique and can't be achieved with the argument.

Update: As we've discussed, we can do it later. I will keep this comment as a reminder for myself to open a new issue once we merge the PR.

packages/cli/src/commands/qualityGate.ts
```diff
@@ -19,50 +21,136 @@ export class QualityGateCommand extends Command {
...
+
+    if (successRate !== undefined) {
+      rules.successRate = successRate;
+    }
+
+    if (fastFail) {
+      rules.fastFail = fastFail;
+    }
+
+    const qualityGateCliRules = {
```
delatrie: Isn't `qualityGateCliRules` the same as `rules`? Can't we use `rules` (perhaps renamed) directly instead?

packages/cli/src/commands/qualityGate.ts
```diff
@@ -19,50 +21,136 @@ export class QualityGateCommand extends Command {
...
 
-    console.error(red(`Quality gate has failed with ${bold(failedResults.length.toString())} errors:\n`));
+    const allureReport = new AllureReport(config);
 
-    for (const result of failedResults) {
-      let scope = "";
+    if (!allureReport.hasQualityGate) {
+      // eslint-disable-next-line no-console
+      console.info("Quality gate is not configured");
+      exit(0);
```
delatrie, line 130: IMO, a misconfigured quality gate shouldn't return the same exit code as a passed one. How about returning -1 in that case?

packages/cli/src/commands/qualityGate.ts
```diff
@@ -19,50 +21,136 @@ export class QualityGateCommand extends Command {
...
+    const state = new QualityGateState();
 
-      switch (result.meta?.type) {
-        case "label":
-          scope = `(label[${result.meta.name}="${result.meta.value}"])`;
-          break;
-        case "parameter":
-          scope = `(parameter[${result.meta.name}="${result.meta.value}"])`;
-          break;
+    allureReport.realtimeSubscriber.onTestResults(async (trsIds) => {
```
delatrie, line 137: WDYT about renaming `realtimeSubscriber`? In the publisher-subscriber model, a subscriber is a recipient of messages (for example, here, the code of the quality gate command is a subscriber).

A proper name would be something like `realtimeEvents`, `realtimeTopic`, `realtimeChannel`, etc.

packages/cli/src/commands/results/pack.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export class ResultsPackCommand extends Command {
+  static paths = [["results", "pack"]];
+
+  static usage = Command.Usage({
+    description: "Creates .zip archive with test results",
+    category: "Allure Test Results",
+    details: "This command creates .zip archive with all test results which can be collected in the project",
+    examples: [
+      ["results pack", "Print information about the current user using the default configuration"],
```
delatrie: This is the description of `whoami`

packages/cli/src/commands/qualityGate.ts
```diff
@@ -19,50 +21,136 @@ export class QualityGateCommand extends Command {
...
+    }
+
+    const qualityGateCliRules = {
+      ...rules,
+    };
+
+    config.plugins = [];
+
+    if (config.qualityGate) {
+      config.qualityGate.rules?.push(qualityGateCliRules);
```
delatrie: Perhaps, we should replace all the configured rules with the one provided via the CLI instead of adding the CLI rule as one extra to the existing ones.

In any case, we shouldn't ignore the CLI args even if the config defines `qualityGate` without `rules` (currently, the CLI rule is ignored because of `?.` operator).

packages/plugin-api/src/index.ts
```diff
@@ -1,8 +1,9 @@
 export * from "./config.js";
 export type * from "./plugin.js";
-export type * from "./qualityGate.js";
 export type * from "./store.js";
 export type * from "./resultFile.js";
+export type * from "./qualityGate.js";
+export type * from "./qualityGate.js";
```
delatrie: Duplicated lines

```suggestion
export type * from "./qualityGate.js";
```

packages/e2e/test/allure-awesome/test/testResult.test.ts
```diff
@@ -96,6 +94,13 @@ test.afterAll(async () => {
 test.describe("allure-awesome", () => {
   test.describe("test results", () => {
     test("it's possible to navigate between tests results using navigation arrows", async () => {
+      await label("sample", "sample0");
+      await label("sample", "sample1");
+      await label("sample", "sample2");
+      await label("sample", "sample3");
+      await label("sample", "sample4");
+      await label("sample", "sample5");
```
delatrie: Leftovers?

packages/core/test/qualityGate/rules.test.ts
```diff
@@ -0,0 +1,283 @@
...
+      knownIssues: [] as KnownTestFailure[],
+      state: 0,
+    });
+
+    expect(result.success).toBe(true);
+    expect(result.actual).toBe(1);
+    expect(result.expected).toBe(expected);
+  });
+
+  // it("should fail when failures count is greater than expected (without state)", async () => {
```
delatrie: What's the status of these commented tests? What further actions are required (uncomment, delete, etc)?

packages/core/src/utils/event.ts
```diff
@@ -15,53 +31,141 @@ interface HandlerData {
...
+  sendTestFixtureResult(tfrId: string) {
+    this.#emitter.emit(RealtimeEvents.TestFixtureResult, tfrId);
+  }
+
+  sendAttachmentFile(afId: string) {
+    this.#emitter.emit(RealtimeEvents.AttachmentFile, afId);
+  }
+}
+
+export class RealtimeSubscriber implements RealtimeSubscriberType {
```
delatrie, line 62: IMO, something like `RealtimeEvents` would be a better name since this class doesn't represent subscribers: callers of this class are the real subscribers.

packages/core/src/report.ts
```diff
@@ -168,6 +176,16 @@ export class AllureReport {
...
+  validate = async (trs: TestResult[], state?: QualityGateState) => {
+    const knownIssues = await this.#store.allKnownIssues();
+
+    return this.#qualityGate!.validate({
+      trs: trs.filter(Boolean),
+      knownIssues,
+      state,
+    });
+  };
+
```
delatrie, line 188: If we only run a quality gate (with no plugins), we can save resources (a huge amount in some workloads) by not loading the history, as none of the rules require it.

Also, I can think of scenarios when appending to the history would also be unexpected. For example, running `allure quality-gate` and `allure generate` separately against the same input.

We can discuss it later and create a separate issue to address it.

packages/core/src/qualityGate/rules.ts
```diff
@@ -0,0 +1,56 @@
+import { filterSuccessful, filterUnsuccessful } from "@allurereport/core-api";
+import { type QualityGateRule } from "@allurereport/plugin-api";
+import { bold } from "yoctocolors";
+
+export const maxFailuresRule: QualityGateRule<number> = {
+  rule: "maxFailures",
+  message: ({ actual, expected }) =>
+    `Maximum number of failed tests ${bold(String(actual))} is more, than expected ${bold(String(expected))}`,
```
delatrie: I find this message confusing. How about changing it a bit? For example:

```suggestion
    `The number of failed tests ${bold(String(actual))} exceeds the allowed threshold value ${bold(String(expected))}`,
```

packages/core/src/qualityGate/rules.ts
```diff
@@ -0,0 +1,56 @@
...
+  validate: async ({ trs, knownIssues, expected, state = 0 }) => {
+    const knownIssuesHistoryIds = knownIssues.map(({ historyId }) => historyId);
+    const unknown = trs.filter((tr) => !tr.historyId || !knownIssuesHistoryIds.includes(tr.historyId));
+    const failedTrs = unknown.filter(filterUnsuccessful);
+    const actual = failedTrs.length + state;
+
+    return {
+      success: actual < expected,
```
delatrie: I would expect a `maxFailures` value to be inclusive:
```suggestion
      success: actual <= expected,
```

packages/core/src/qualityGate/rules.ts
```diff
@@ -0,0 +1,56 @@
...
+      actual,
+      expected,
+    };
+  },
+};
+
+export const minTestsCountRule: QualityGateRule<number> = {
+  rule: "minTestsCount",
+  message: ({ actual, expected }) =>
+    `Minimum number of tests ${bold(String(actual))} is less, than expected ${bold(String(expected))}`,
```
delatrie: I find this message confusing as well. We can rephrase it like this:

```suggestion
    `The total number of tests ${bold(String(actual))} is less than the expected threshold value ${bold(String(expected))}`,
```

packages/cli/src/commands/run.ts
```diff
@@ -81,22 +101,73 @@ const runTests = async (
...
+      if (!fastFailed) {
+        return;
+      }
+
+      allureReport.realtimeDispatcher.sendQualityGateResult(results);
+      qualityGateResults = results;
+
+      try {
+        // @ts-ignore
+        await terminate(testProcess.pid);
```
delatrie: `terminate` sends `SIGKILL` by default, which is quite brutal.
Let's use a less violent method instead. A conventional signal for termination requests is `SIGTERM`:

```suggestion
        await terminate(testProcess.pid, "SIGTERM");
```

packages/cli/src/commands/run.ts
```diff
@@ -81,22 +101,73 @@ const runTests = async (
...
+        throw err;
+      }
+    });
   }
 
-  await processWatcher.abort();
-  return code;
+  return {
+    process: testProcess,
+    // FIXME: at this moment, this is the cheapest way to have the process to control and keep ability to receive original exit code
```
delatrie: IMO, the way you did this is good, and this comment can be removed. It's up to you, though

packages/cli/src/commands/results/pack.ts
```diff
@@ -0,0 +1,123 @@
...
+      ],
+    ],
+  });
+
+  resultsDir = Option.String({
+    required: false,
+    name: "The directory with Allure results. Entire working directory will be scanned for `allure-results` unless this option is provided.",
+  });
+
+  pattern = Option.String("--pattern", {
```
delatrie: I still believe that using glob patterns for result directories would be a better solution:

```
 yarn allure results pack ./packages/*/out/allure-results/
 yarn allure results pack ./**/allure-results/
```

It simplifies the whole thing, making the `--pattern` parameter redundant. In many cases, it's also much faster compared to the full recursive scan of the CWD, which is what we currently perform by default.

However, let's at least rename `--pattern` to something like `--results-dir-name` or similar, because people usually expect support for a special syntax in patterns (like `*` and `**` in glob patterns), while in reality we're only using the value for direct comparison with the candidate directory names.
Given the description, users will likely be frustrated by the inability to use things like `--pattern ./packages/*/out/allure-results`

delatrie (changes requested):

packages/core/src/qualityGate/rules.ts
delatrie: BTW, should we count retries? We currently do, but my impression is that we shouldn't.
If `minTestsCount` is 2, and we have a single test with 3 retries, the quality gate doesn't seem to pass to me. The same goes for `maxFailures`.
The same question goes about environments.

packages/plugin-api/src/charts.ts
```diff
@@ -316,33 +316,28 @@ export const generatePieChart = (
...
   getHistoricalData: (historyPoint: HistoryDataPoint) => TrendStats<T>;
   // List of all possible values for the type
   getAllValues: () => readonly T[];
 }
 
-export const generateTrendChartGeneric = async <T extends TrendDataType>(
+export const generateTrendChartGeneric = <T extends TrendDataType>(
   options: TrendChartOptions,
-  store: AllureStore,
+  trs: TestResult[] = [],
```
Viperwow: I see the idea behind it - to minimise async calls to the store. However, let's make its form similar to getPieChartData, which uses an object with stored data (`{ statistic: ... }`). It will be easier to operate with if more data comes out to be used, and easier to deal with merge conflicts :)

packages/cli/src/commands/results/pack.ts
epszaw: Done

packages/cli/src/commands/results/pack.ts
epszaw: Done

packages/cli/src/commands/generate.ts
epszaw: Done

packages/cli/src/commands/qualityGate.ts
epszaw: Agree. Done

packages/cli/src/commands/qualityGate.ts
epszaw: Already was solved

packages/cli/src/commands/qualityGate.ts
epszaw: Already solved

packages/cli/src/commands/qualityGate.ts
epszaw: Let's leave it as is for now and rename in next iterations

packages/cli/src/commands/run.ts
epszaw: Ok :)

packages/cli/src/commands/run.ts
epszaw: :+1:

packages/core/src/report.ts
epszaw: Let's create issue for that

packages/core/src/utils/event.ts
epszaw: As I wrote – let's rename it in next iterations

packages/cli/src/commands/run.ts
epszaw: In case of the process termination, we won't see the message, so I added similar message in the termination branch

packages/plugin-api/src/charts.ts
epszaw: :+1:

packages/cli/src/commands/generate.ts
```diff
@@ -39,16 +45,43 @@ export class GenerateCommand extends Command {
...
       output: this.output ?? "allure-report",
     });
+    const matcher = pm(resultsDir, {
+      dot: true,
+      contains: true,
+    });
+    const resultsDirectories = new Set<string>();
+
+    await findMatching(targetDir, resultsDirectories, (dirent) => {
+      if (dirent.isFile()) {
```
delatrie: Better use `!dirent.isDirectory()`.

`isFile` returns `false` for, say, symlinks, which we may hit upon when scanning the full repo.

packages/cli/src/commands/qualityGate.ts
```diff
@@ -19,50 +23,142 @@ export class QualityGateCommand extends Command {
...
+      dot: true,
+      contains: true,
+    });
 
-    await allureReport.start();
-    await allureReport.readDirectory(this.resultsDir);
-    await allureReport.done();
-    await allureReport.validate();
+    await findMatching(cwd, resultsDirectories, (dirent) => {
+      if (dirent.isFile()) {
```
delatrie: Same as in generate.ts: better use `!dirent.isDirectory()`

packages/cli/src/commands/results/pack.ts
```diff
@@ -0,0 +1,128 @@
...
+    const archiveName = this.name ?? "allure-results.zip";
+    const resultsDirectories = new Set<string>();
+    const resultsFiles = new Set<string>();
+    const matcher = pm(resultsDir, {
+      dot: true,
+      contains: true,
+    });
+
+    await findMatching(cwd, resultsDirectories, (dirent) => {
+      if (dirent.isFile()) {
```
delatrie: Same as in generate.ts: better use `!dirent.isDirectory()`

packages/cli/src/commands/run.ts
```diff
@@ -81,22 +101,76 @@ const runTests = async (
...
+
+        throw err;
+      }
+    });
   }
 
-  await processWatcher.abort();
-  return code;
+  return {
+    process: testProcess,
```
delatrie: The `process` property isn't used, and all calls to `runTests` are immediately followed by `await testProcess.processTerminationPromise;`, which means you can return `await new Promise(...)` and remove an extra await

packages/cli/src/commands/run.ts
delatrie: > In case of the process termination, we won't see the message

I'm not sure that's the case. Here I run Playwright tests with `maxFailures=0` and `fastFail=true`. When `allure run` hits a failed test, it terminates Playwright and prints the message `process finished with code 0 (17614)ms`:

```
> yarn allure run --silent -- yarn run playwright test ./test/allure-awesome/features/retries.test.ts --project chromium
/Users/maksimstepanov/repos/allure-framework/allure3/packages/e2e
yarn run playwright test ./test/allure-awesome/features/retries.test.ts --project chromium
process finished with code 0 (17614)ms
Quality Gate failed with following issues:

 ⨯ The number of failed tests 1 exceeds the allowed threshold value 0    maxFailures

1 quality gate rules have been failed.
exit code 1 (17680ms)
```

So the message is shown just fine.

The problem is that the message doesn't tell us the truth: the process hasn't exited with code 0; it has exited by `SIGTERM`. I'm not sure what exit code Playwright uses in such a case, but it's definitely not 0 (Node.js doesn't give us the API to get it).

Two better options come into my mind:

  - Option 1: Get `signal` in the same way we get `code` and check it. If it's not `null`, then print another message.
  - Option 2: If `code` is `null`, print something like `process was terminated (100500ms)`

packages/cli/src/commands/run.ts
```diff
@@ -81,22 +101,76 @@ const runTests = async (
...
+        const afterProcess = Date.now();
 
-  await allureResultsWatch.abort();
+        // @ts-ignore
+        await terminate(testProcess.pid, "SIGTERM");
 
-  for (const [ar, watcher] of allureResultsWatchers) {
-    await watcher.abort();
-    allureResultsWatchers.delete(ar);
+        console.log(`process terminated with code 1 (${afterProcess - beforeProcess})ms`);
```
delatrie: We won't see this message because `terminate` leads to `on('exit', ...)` which, in turn, calls `system.exit`. In such a case, the rest of the function won't be called.

I think we can remove it because we already have a `process finished` message, which is displayed in all cases (I tested that on my machine).

packages/cli/src/commands/run.ts
```diff
@@ -209,28 +293,60 @@ export class RunCommand extends Command {
...
+
         logTests(await allureReport.store.allTestResults());
       }
 
       await allureReport.done();
-      await allureReport.validate();
 
-      process.exit(code ?? allureReport.exitCode);
+      if (!withQualityGate) {
+        process.exit(testProcessResult?.code ?? 0);
```
delatrie: Better to return a non-zero exit code if `code` is `null`, which in Node.js means the process has been terminated by a signal. It's not a successful execution (for example, Playwright exits with code 143 if it gets `SIGTERM`).

```suggestion
        process.exit(testProcessResult?.code ?? -1);
```

packages/cli/src/commands/run.ts
```diff
@@ -209,28 +293,60 @@ export class RunCommand extends Command {
 
         const tmpDir = await mkdtemp(join(tmpdir(), "allure-run-"));
         const testPlanPath = resolve(tmpDir, `${rerun}-testplan.json`);
+
         await writeFile(testPlanPath, JSON.stringify(testPlan));
 
-        code = await runTests(
+        testProcess = await runTests({
```
delatrie: `allure run` won't work properly if there is a `fastFail` rule and `--rerun` is greater than 0.
I suggest exiting early in such a case and printing a message explaining that until we come up with a solution.

I will open an issue after we merge the PR.

packages/core/src/qualityGate/rules.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const maxFailuresRule: QualityGateRule<number> = {
+  rule: "maxFailures",
+  message: ({ actual, expected }) =>
+    `The number of failed tests ${bold(String(actual))} exceeds the allowed threshold value ${bold(String(expected))}`,
+  validate: async ({ trs, knownIssues, expected, state = 0 }) => {
+    const knownIssuesHistoryIds = knownIssues.map(({ historyId }) => historyId);
+    const unknown = trs.filter((tr) => !tr.historyId || !knownIssuesHistoryIds.includes(tr.historyId));
+    const failedTrs = unknown.filter(filterUnsuccessful);
+    const actual = failedTrs.length + state;
```
delatrie, line 13: All rules must correctly account for retries. In other words, they must count unique `historyId` values, not test result objects.

As we've discussed, this will be the subject of a separate PR, so I will create a new issue once we merge this PR.

> [!NOTE]
> Currently, this only affects the `maxFailures` rule with `fastFail`. Other rules don't make any sense with `fastFail`. And if `fastFail` is false, `allure run` will apply QG at the end of the run, where all retries are correctly grouped under the most recent result objects.

packages/cli/src/commands/generate.ts
```diff
@@ -39,16 +45,43 @@ export class GenerateCommand extends Command {
...
       name: this.reportName,
       output: this.output ?? "allure-report",
     });
+    const matcher = pm(resultsDir, {
+      dot: true,
+      contains: true,
+    });
+    const resultsDirectories = new Set<string>();
+
+    await findMatching(targetDir, resultsDirectories, (dirent) => {
```
delatrie: We should pass `cwd` instead of `targetDir`. Otherwise:

  - for `allure generate ./my-results`, the directory won't be enumerated by `findMatching`
  - for `allure generate ./packages/*/out/allure-results`, `findMatching` won't find anything because `opendir` will fail with `ENOENT`.

BTW, `pack` and `quality-gate` pass `cwd` to `findMatching`, which seems correct and doesn't lead to the abovementioned errors.

epszaw: Good catch, thank you.

packages/cli/src/commands/generate.ts
epszaw: Done

packages/cli/src/commands/results/pack.ts
epszaw: Done

packages/cli/src/commands/run.ts
epszaw: Done

packages/cli/src/commands/run.ts
epszaw: The second option is good. Added, resolved

packages/cli/src/commands/run.ts
epszaw: Done

packages/cli/src/commands/run.ts
epszaw: Added notification

packages/core/src/qualityGate/rules.ts
epszaw: We can simply pass not-hidden test results on the upper layer. Fixed

packages/cli/src/commands/qualityGate.ts
epszaw: Done

packages/core/src/report.ts
epszaw: I added ability to pass known issues as a parameter. No need to create issue

packages/core/src/qualityGate/rules.ts
epszaw: Done

packages/cli/src/commands/run.ts
epszaw: You're right. Removed the property

packages/cli/src/commands/run.ts
```diff
@@ -81,22 +97,74 @@ const runTests = async (
...
+      }
+    });
+  }
+
   const code = await terminationOf(testProcess);
   const afterProcess = Date.now();
 
-  console.log(`process finished with code ${code ?? 0} (${afterProcess - beforeProcess})ms`);
+  if (code !== null) {
+    console.log(`process finished with code ${code} (${afterProcess - beforeProcess})ms`);
```
delatrie:
```suggestion
    console.log(`process finished with code ${code} (${afterProcess - beforeProcess}ms)`);
```

packages/cli/src/commands/run.ts
```diff
@@ -81,22 +97,74 @@ const runTests = async (
...
+  } else {
+    console.log(`process terminated (${afterProcess - beforeProcess})ms`);
```
delatrie:
```suggestion
    console.log(`process terminated (${afterProcess - beforeProcess}ms)`);
```

packages/cli/src/commands/generate.ts
```diff
@@ -39,16 +45,43 @@ export class GenerateCommand extends Command {
...
 
   async execute() {
-    const config = await readConfig(this.cwd, this.config, {
+    const cwd = this.cwd ?? process.cwd();
+    const resultsDir = this.resultsDir ?? "./**/allure-results";
+    const config = await readConfig(cwd, this.config, {
       name: this.reportName,
       output: this.output ?? "allure-report",
     });
+    const matcher = pm(resultsDir, {
```
delatrie, line 54: It's sensitive to trailing slashes:

1. `yarn allure generate ./packages/e2e/out/allure-results/` - finds nothing
2. `yarn allure generate ./packages/e2e/out/allure-results` - works

Trailing slashes are often added automatically by the shell when using autocompletion, so we'd better remove the trailing `/` (and `\` on Windows). The same goes for `pack` and `quality-gate`.

epszaw: Solved

PR #265, opened 2025-07-22
fix/#117-fix-indicators-weight

Viperwow:
- Add tooltips to `transition` statuses
- Change `Tags` to just icons in the `transition` statuses to simplify visuals and make it less "alerting"
- Remove old logic related to unnecessary translations


Viperwow:
Stale one. Check https://github.com/allure-framework/allure3/pull/266

PR #269, opened 2025-07-24
fix(awesome): the "Retries" tab of a test result without start/stop timestamps breaks the report

delatrie:
An entry on the Retries tab has a title that includes the formatted stop timestamp of the corresponding test result. However, some readers produce test results that lack timestamps. For example, the XCResult format doesn't include timestamps at the test result level.

In such cases, opening the "Retries" tab causes the report to become broken: it becomes impossible to navigate, and the content pane becomes empty and collapsed. Here is an example (see the report [here](https://delatrie.github.io/allure-xcresult-samples/repetitions/v3/)).

image

The PR fixes that by omitting the timestamp text in case `stop` is zero.

#### Improved retry titles

Additionally, it prefixes the timestamp with the following text: `Attempt {{number}} of {{total}}`. The text serves multiple purposes:

1. It prevents entries from being empty in case the timestamps are missing.
2. It makes it easier to figure out how retries are sorted, especially in case the provided resolution (which is 1 second) isn't sufficient to tell apart the retries.

#### Retry fixture and tests

The PR introduces the `RetryItemFixture` class to help with validating the behavior of the Retries tab entries. Use `TestResultPage.getRetry` to access it.

Also, the PR adds a couple of tests that cover retry titles. Ideally, more tests are required to cover buttons, details, and other aspects.

#### Port to classic

The PR also ports those changes to `web-classic`, although it's not possible to verify them since the classic report currently doesn't work.

### DIFF
--- a/packages/web-awesome/src/components/TestResult/TrRetriesView/TrRetriesItem.tsx
+++ b/packages/web-awesome/src/components/TestResult/TrRetriesView/TrRetriesItem.tsx
@@ -29,14 +29,20 @@ export const TrRetriesItem: FunctionalComponent<TrRetriesItemProps> = ({ testRes
   const navigateUrl = id;
 
   return (
-    <div>
+    <div data-testid="test-result-retries-item">
       <div className={styles["test-result-retries-item-header"]} onClick={() => setIsOpen(!isOpened)}>
         {Boolean(error.trace || error.message) && (
-          <ArrowButton isOpened={isOpened} icon={allureIcons.lineArrowsChevronDown} />
+          <ArrowButton
+            data-testid="test-result-retries-item-arrow-button"
+            isOpened={isOpened}
+            icon={allureIcons.lineArrowsChevronDown}
+          />
         )}
         <div className={styles["test-result-retries-item-wrap"]}>
           <TreeItemIcon status={status} className={styles["test-result-retries-item-status"]} />
-          <Text className={styles["test-result-retries-item-text"]}>{retryTitle}</Text>
+          <Text data-testid="test-result-retries-item-text" className={styles["test-result-retries-item-text"]}>
+            {retryTitle}
+          </Text>
           <div className={styles["test-result-retries-item-info"]}>
             {Boolean(formattedDuration) && (
               <Text type="ui" size={"s"} className={styles["item-time"]}>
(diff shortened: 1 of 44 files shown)


packages/web-awesome/src/components/TestResult/TrRetriesView/TrRetriesItem.tsx
```diff
@@ -5,27 +5,44 @@ import { useState } from "preact/hooks";
...
+export const TrRetriesItem: FunctionalComponent<TrRetriesItemProps> = ({ testResultItem, attempt, totalAttempts }) => {
   const { id, status, error, stop, duration } = testResultItem;
   const [isOpened, setIsOpen] = useState(false);
-  const convertedStop = timestampToDate(stop);
+
+  const { t } = useI18n("ui");
+
+  const retryTitlePrefix = t("attempt", { attempt, total: totalAttempts });
+  const convertedStop = stop ? timestampToDate(stop) : undefined;
+  const retryTitle = convertedStop ? `${retryTitlePrefix} -- ${convertedStop}` : retryTitlePrefix;
```
epszaw:
```suggestion
  const retryTitle = convertedStop ? `${retryTitlePrefix} – ${convertedStop}` : retryTitlePrefix;
```

packages/web-classic/src/components/TestResult/TestResultRetriesView/TestResultRetriesItem.tsx
```diff
@@ -6,16 +6,30 @@ import { ArrowButton } from "@/components/ArrowButton";
...
+}) => {
   const { id, status, message, trace, stop, duration } = testResultItem;
   const [isOpened, setIsOpen] = useState(false);
-  const convertedStop = timestampToDate(stop);
-  const formattedDuration = formatDuration(duration as number);
+  const { t } = useI18n("ui");
+
+  const retryTitlePrefix = t("attempt", { attempt, total: totalAttempts });
+  const convertedStop = stop ? timestampToDate(stop) : undefined;
+  const retryTitle = convertedStop ? `${retryTitlePrefix} -- ${convertedStop}` : retryTitlePrefix;
```
epszaw:
```suggestion
  const retryTitle = convertedStop ? `${retryTitlePrefix} – ${convertedStop}` : retryTitlePrefix;
```

packages/web-awesome/src/components/TestResult/TrRetriesView/TrRetriesItem.tsx
delatrie: Agree, dashes look better. I applied your suggestions.

image

PR #277, opened 2025-08-08
Fix cli flags re-assignment

epszaw:


packages/core/src/config.ts
```diff
@@ -156,23 +156,16 @@ export const getPluginInstance = (config: FullConfig, predicate: (plugin: Plugin
 
 /**
  * Enforces the plugin instance in the config
- * If the plugin instance is not present, it will be added to the config as a single plugin in it
- * If the plugin instance is already present, it will be used as a single plugin in it
+ * Fully replaces the existing plugin in the config or add it, when it doesn't exist
```
delatrie: IIUC, you've removed all `enforcePlugin` calls. How about we get rid of this function entirely?

delatrie (approved):

PR #293, opened 2025-08-22
Add more report data (global errors, global exit code, quality gate results, etc.)

epszaw:
add ability to extract stdout/stderr from test process (and attach them in future)

### DIFF
--- a/packages/e2e/test/allure-awesome/features/globals.test.ts
+++ b/packages/e2e/test/allure-awesome/features/globals.test.ts
@@ -38,7 +38,7 @@ test.describe("globals", () => {
 
       await page.goto(bootstrap.url);
 
-      await expect(page.getByTestId("report-data")).toContainText("with code 1");
+      await expect(page.getByTestId("report-data")).toContainText("with exit code 1");
       await globalsPage.attachScreenshot();
     });
 
@@ -51,7 +51,7 @@ test.describe("globals", () => {
         testResults: [],
         globals: {
           exitCode: {
-            actual: 2,
+            actual: 0,
             original: 1,
           },
         },
@@ -59,7 +59,7 @@ test.describe("globals", () => {
 
       await page.goto(bootstrap.url);
 
-      await expect(page.getByTestId("report-data")).toContainText("with code 2 (original 1)");
+      await expect(page.getByTestId("report-data")).toContainText("with exit code 0 (original 1)");
       await globalsPage.attachScreenshot();
     });
   });


packages/cli/src/commands/run.ts
```diff
@@ -140,6 +144,29 @@ const runTests = async (params: {
     });
   }
 
+  testProcess.stdout?.on?.("data", (data: Buffer) => {
+    const chunk = data.toString("utf8");
```
delatrie: `utf-8` is a multibyte encoding. `setEncoding` is mandatory in such cases:

```js
testProcess.stdout?.setEncoding("utf-8").on?.("data", (chunk: string) => {
  // ...
});
```

Otherwise, the data may get corrupted because raw `on("data")` (without encoding) doesn't guarantee chunks to be aligned with code points encoded in UTF-8.

Alternative: don't convert `Buffer` to a string until we get the entire output and pass it to `process.stdout.write` instead of `console.log`:

```js
testProcess.stdout?.on?.("data", (chunk: Buffer) => {
  testProcessStdoutRaw = Buffer.concat([testProcessStdoutRaw, chunk]);
  if (!silent) {
    process.stdout.write(chunk);
  }
});
const stdout = testProcessStdoutRaw.toString("utf-8");
```

It spares one extra decoding but is slightly more complex.

packages/cli/src/commands/run.ts
```diff
@@ -316,50 +352,70 @@ export class RunCommand extends Command {
...
-      if (!results.length) {
-        exit(0);
-        return;
+      if (withQualityGate) {
+        globalExitCode.actual = qualityGateResults.length > 0 ? 1 : 0;
       }
 
-      allureReport.realtimeDispatcher.sendQualityGateResult(results);
-      qualityGateMessage = stringifyQualityGateResults(results);
+      if (testProcessResult?.stderr) {
```
delatrie: Some programs use stderr to write logs/warnings/diagnostic messages.

Example from our own repo:

```
yarn allure run --cwd ./packages/reader  --config ../../allurerc.mjs -- yarn run vitest run ./test/allure2.test.ts > /dev/null
```

The exit code is zero, and all tests pass, but the stderr is not empty:

```
Sourcemap for "/.../allure3/.yarn/__virtual__/allure-vitest-virtual-3a0053d30e/0/cache/allure-vitest-npm-3.3.3-9424586b36-df1e4506cb.zip/node_modules/allure-vitest/dist/setup.js" points to missing source files
```

Better add a global error based on the exit code and attach stderr separately (if it's present).

packages/web-awesome/src/locales/en.json
```diff
@@ -119,7 +122,9 @@
     "attempt": "Attempt {{attempt}} of {{total}}",
     "at": "at",
     "variables": "Variables",
-    "openPwTrace": "Open Playwright Trace"
+    "openPwTrace": "Open Playwright Trace",
+    "finishedAtOriginal": "{{formattedCreatedAt}}, with code {{original}}",
```
delatrie: It may not be obvious to some people that we're talking about the exit codes in a phrase like `August 27, 2025 at 11:10:00 PM, with code 1 (original 1)`. How about keeping the term "exit code" in full? Maybe just `exit code` instead of `with code`?

delatrie (approved):

packages/web-awesome/src/locales/en.json
epszaw: Good idea

packages/cli/src/commands/run.ts
epszaw: Good catch! Fixed

packages/cli/src/commands/run.ts
epszaw: Agree. Done

PR #295, opened 2025-08-26
Feat/#190 add bar chart

Viperwow:


packages/web-components/stories/BarChart/mocks.ts
```diff
@@ -0,0 +1,76 @@
+import type { TestStatus, SeverityLevel } from "@allurereport/core-api";
+
+// Константы для статусов тестов
```
epszaw: Please, use English in the comments

PR #307, opened 2025-09-16
Feat/#194 add tree map

Viperwow:


Viperwow:
Need to fix:
- [x] Inconsistent behaviour for a single test out of the label group
- [x] Inconsistent behaviour for a single test in the mixed group node on parent value calculation (huge test base scenario)
- [x] Group leaf label not being hidden on block being smaller than the threshold
- [ ] Add TreeMapChart tooltips highlighting all provided data in a Tree leaf

packages/web-components/src/components/Widget/index.tsx
```diff
@@ -10,7 +10,6 @@ export const Widget: FunctionalComponent<WidgetProps> = ({ children, title }) =>
   return (
     <div className={styles.widget}>
       <div className={styles.header}>
-        <div className={styles.dragArea} />
```
epszaw, line 13: Why did you decide to remove the handler?

PR #309, opened 2025-09-19
Multi-stage reports

epszaw:


packages/cli/src/commands/generate.ts
```diff
@@ -20,6 +21,10 @@ export class GenerateCommand extends Command {
         "generate ./allure-results --output custom-report",
         "Generate a report from the ./allure-results directory to the custom-report directory",
       ],
+      [
+        "generate --stage=windows --stage=macos.zip ./allure-results",
```
delatrie:
```suggestion
        "generate --stage=windows.zip --stage=macos.zip ./allure-results",
```

packages/cli/src/commands/run.ts
```diff
@@ -210,6 +210,10 @@ export class RunCommand extends Command {
     examples: [
       ["run -- npm run test", "Run npm run test and collect Allure results"],
       ["run --rerun 3 -- npm run test", "Run npm run test and rerun failed tests up to 3 times"],
+      [
+        "run --stage=my-stage -- npm run test",
```
delatrie, line 214: A minor suggestion:
WDYT about making `--stage` a flag and reuse `--output` for a file name? After we change `--stage` of `generate` to be a pattern, the symmetry between run and generate isn't that apparent anymore.

```shell
# Save to a specific file:
npx allure run --stage --output ./AllureStage-Linux-Chrome.zip -- npm test

# Save to a default file (like allure-stage.zip; can then be uploaded as an artifact with a proper name):
npx allure run --stage -- npm test
```

packages/core/src/report.ts
```diff
@@ -243,6 +381,10 @@ export class AllureReport {
     // closing it early, to prevent future reads
     this.#stage = "done";
 
+    if (this.#transitionalStage) {
+      await this.dumpState();
+    }
+
```
delatrie, line 463: We shouldn't generate reports if `--stage` is provided (currently, we do). Should we skip all the plugins or only those that generate files is an open question for me.

UPDATE: I've checked the reworked version, and `allure run --stage ...` still generates the reports.

packages/core/src/report.ts
```diff
@@ -231,6 +249,126 @@ export class AllureReport {
...
+          await writeFile(attachmentFilePath, attachmentContentEntry);
+
+          resultsAttachments[attachmentId] = new PathResultFile(attachmentFilePath);
+        }
+      } catch (err) {
+        console.error(`Can't restore state from "${stage}", continuing without it`);
+        console.error(err);
+      } finally {
+        await rm(stageTempDir, { recursive: true });
```
delatrie: We can't remove the files until the plugins complete their job. At this point, no `update` or `done` plugin methods have been called. We should keep the files until `done` is completed for all plugins.

packages/cli/src/commands/generate.ts
```diff
@@ -41,33 +50,62 @@ export class GenerateCommand extends Command {
...
     const resultsDirectories = new Set<string>();
 
-    await findMatching(cwd, resultsDirectories, (dirent) => {
-      if (dirent.isDirectory()) {
-        const fullPath = join(dirent?.parentPath ?? dirent?.path, dirent.name);
+    if (this.stage?.length) {
+      for (const stage of this.stage) {
+        const matcher = pm(stage, {
+          dot: true,
+          contains: true,
```
delatrie: We shouldn't use `contains: true` as it results in unexpected matches.

Imagine:

```
$ tree
.
├── docs
│   └── how_to_get_allure-stage.zip.md
└── allure-stage.zip

$ npx allure generate --stage allure-stage
```

The command will attempt to restore from `allure-stage.zip` and `docs/how_to_get_allure-stage.zip.md`. It's not expected for a pattern without directory separators to find files inside directories.

The correct way would be to set `contains` to `false` and check relative paths instead of absolute:

```diff
- return matcher(fullPath);
+ return matcher(relative(cwd, fullPath));
```

I believe this should also apply to allure-results patterns.

packages/core/src/report.ts
```diff
@@ -231,6 +249,126 @@ export class AllureReport {
...
+
+    for (const attachment of allAttachments) {
+      const content = await this.#store.attachmentContentById(attachment.id);
+
+      if (!content) {
+        continue;
+      }
+
+      if (content instanceof PathResultFile) {
+        await addEntry(content.path, {
```
delatrie:
```suggestion
        await addEntry(createReadStream(content.path), {
```

packages/core/src/store/store.ts
```diff
@@ -610,4 +655,29 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
+
+    mergeMapWithRecord(this.#testResults, testResults);
+    mergeMapWithRecord(this.#attachments, attachments);
+    mergeMapWithRecord(this.#testCases, testCases);
+    mergeMapWithRecord(this.#fixtures, fixtures);
+
+    this.#addEnvironments(environments);
+
+    Object.assign(this.#reportVariables, reportVariables);
+    Object.assign(this.#attachmentContents, attachmentsContents);
```
delatrie: This seems not to be enough for attachments to work. I observed the attachments directory missing in `allure-report/awesome/data`, and the attachments, while being present on the test result pages, appeared empty.

Unfortunately, I can't check this after the rework because the whole report is now not generating properly.

delatrie (changes requested):
After the rework, I can't generate the report from stage data anymore:

1. `yarn allure run --stage=my-stage --environment=my-env -- yarn workspace @allurereport/reader run vitest run ./test/allure2.test.ts`
2. `yarn allure generate --stage my-stage.zip`
3. `yarn open ./allure-report`
4. Open "Allure Awesome"

image

packages/core/src/report.ts
epszaw: You're right. Done

packages/cli/src/commands/generate.ts
epszaw: Good catch 👍

packages/cli/src/commands/generate.ts
epszaw: I did it intentionally to show, that it's possible to omit `.zip` extension

packages/core/src/report.ts
epszaw: Fixed

packages/cli/src/commands/generate.ts
```diff
@@ -41,33 +49,50 @@ export class GenerateCommand extends Command {
...
-      if (dirent.isDirectory()) {
-        const fullPath = join(dirent?.parentPath ?? dirent?.path, dirent.name);
-
-        return matcher(fullPath);
+    const stageDumpFiles: string[] = [];
+    const resultsDirectories: string[] = [];
+
+    if (this.stage?.length) {
+      for (const stage of this.stage) {
+        const globPattern = join(cwd, stage);
```
delatrie: The result will contain `\` as a directory separator on Windows. However, glob uses `\` for escaping and requires the use of `/` as path separators on all platforms. Otherwise, no results will be returned on Windows.

Use `cwd` and `absolute` options instead:

```js
const matchedFiles = await glob(stage, {
  nodir: true,
  dot: true,
  absolute: true,
  cwd,
});
```

packages/cli/src/commands/generate.ts
```diff
@@ -41,33 +49,50 @@ export class GenerateCommand extends Command {
...
+        stageDumpFiles.push(...matchedFiles);
       }
+    }
 
-      return false;
-    });
+    // don't read allure results directories without the parameter when stage file has been found
+    // or read allure results directory when it is explicitly provided
+    if (!!this.resultsDir || stageDumpFiles.length === 0) {
+      const globPattern = join(cwd, resultsDir ?? "./**/allure-results");
```
delatrie: Use `cwd` and `absolute` options of `glob` instead of `path.join` (see above).

packages/cli/src/commands/generate.ts
```diff
@@ -41,33 +49,50 @@ export class GenerateCommand extends Command {
...
+      const matchedDirs = (
+        await glob(globPattern, {
+          mark: true,
+          nodir: false,
+        })
+      ).filter((p) => p.endsWith("/"));
```
delatrie: On Windows, directory paths end with `\\`. Use `path.sep` instead of a literal.

packages/cli/src/commands/qualityGate.ts
```diff
@@ -114,17 +108,15 @@ export class QualityGateCommand extends Command {
...
-    await findMatching(cwd, resultsDirectories, (dirent) => {
-      if (dirent.isDirectory()) {
-        const fullPath = join(dirent?.parentPath ?? dirent?.path, dirent.name);
-
-        return matcher(fullPath);
-      }
-
-      return false;
-    });
+    const globPattern = join(cwd, resultsDir);
```
delatrie: Use `cwd` and `absolute` options of `glob` instead of `path.join` (see above).

packages/cli/src/commands/qualityGate.ts
```diff
@@ -114,17 +108,15 @@ export class QualityGateCommand extends Command {
...
+    const resultsDirectories = (
+      await glob(globPattern, {
+        mark: true,
+        nodir: false,
+      })
+    ).filter((p) => p.endsWith("/"));
```
delatrie: Use `path.sep` (see above)

packages/cli/src/commands/results/pack.ts
```diff
@@ -62,24 +61,16 @@ export class ResultsPackCommand extends Command {
     const cwd = await realpath(this.cwd ?? process.cwd());
     const resultsDir = (this.resultsDir ?? "./**/allure-results").replace(/[\\/]$/, "");
     const archiveName = this.name ?? "allure-results.zip";
-    const resultsDirectories = new Set<string>();
+    const globPattern = join(cwd, resultsDir);
```
delatrie: Use `cwd` and `absolute` options of `glob` instead of `path.join` (see above).

packages/cli/src/commands/results/pack.ts
```diff
@@ -62,24 +61,16 @@ export class ResultsPackCommand extends Command {
...
+    const resultsDirectories = (
+      await glob(globPattern, {
+        mark: true,
+        nodir: false,
+      })
+    ).filter((p) => p.endsWith("/"));
```
delatrie: Use `path.sep` (see above)

packages/cli/src/commands/generate.ts
delatrie: `windows` doesn't match against `windows.zip` (`allure run` always adds `.zip` at the end). The exact match or `*` is required.

packages/core/src/report.ts
```diff
@@ -231,6 +250,153 @@ export class AllureReport {
...
+        testCases: JSON.parse(testCasesEntry.toString("utf8")),
+        fixtures: JSON.parse(fixturesEntry.toString("utf8")),
+        attachments: JSON.parse(attachmentsEntry.toString("utf8")),
+        environments: JSON.parse(environmentsEntry.toString("utf8")),
+        reportVariables: JSON.parse(reportVariablesEntry.toString("utf8")),
+        globalAttachments: JSON.parse(globalAttachmentsEntry.toString("utf8")),
+        globalErrors: JSON.parse(globalErrorsEntry.toString("utf8")),
+        indexAttachmentByTestResult: JSON.parse(indexAttachmentsEntry.toString("utf8")),
+      };
+      const stageTempDir = await mkdtemp(stage);
```
delatrie: Let's create temp directories in `tmpdir()` instead of cwd. Otherwise, IDEs might get crazy in case many attachments are unpacked:

```js
const stageTempDir = await mkdtemp(join(tmpdir(), stage));
```

packages/plugin-api/src/config.ts
```diff
@@ -8,6 +8,16 @@ export interface Config {
   historyPath?: string;
   knownIssuesPath?: string;
   defaultLabels?: DefaultLabelsConfig;
+  /**
+   * Signals that the report's plugins shouldn't be executed, but test results should be archived
+   * Archived test results can be restored later
+   */
+  transitionalStage?: string;
```
delatrie: a nitty question, but why transitional?
Shouldn't we keep the symmetry between the config and the CLI args and call it just `stage` as well? WDYT?

epszaw: Why not? I called the property `transitionalStage` because we already have `stage` in the report, so now, I renamed `transitionalStage` to `stage` and the `#stage` report's property to `#executionStage`

packages/core/src/report.ts
epszaw: Done

packages/cli/src/commands/results/pack.ts
epszaw: Done

packages/cli/src/commands/qualityGate.ts
epszaw: Done

packages/cli/src/commands/generate.ts
epszaw: Done

packages/cli/src/commands/generate.ts
epszaw: Done

packages/cli/src/commands/generate.ts
epszaw: Done

packages/cli/src/commands/qualityGate.ts
epszaw: Done

packages/cli/src/commands/results/pack.ts
epszaw: Done

packages/core/src/report.ts
```diff
@@ -223,25 +243,180 @@ export class AllureReport {
...
+        environments: JSON.parse(environmentsEntry.toString("utf8")),
+        reportVariables: JSON.parse(reportVariablesEntry.toString("utf8")),
+        globalAttachments: JSON.parse(globalAttachmentsEntry.toString("utf8")),
+        globalErrors: JSON.parse(globalErrorsEntry.toString("utf8")),
+        indexAttachmentByTestResult: JSON.parse(indexAttachmentsEntry.toString("utf8")),
+      };
+      const stageTempDir = join(tmpdir(), stage);
+      const resultsAttachments: Record<string, ResultFile> = {};
+
+      await mkdir(stageTempDir, { recursive: true });
```
delatrie: I would switch back to `mkdtemp`, which adds random characters to the directory name. For example, this call:

```js
const stageTempDir = await mkdtemp(path.join(tmpdir(), "windows-"));
```

will create a directory like this: `C:\\Users\\username\\AppData\\Local\\Temp\\windows-ztTUGa` (it's on Windows).

Otherwise, two allure3 runs may interfere with each other.

packages/core/src/report.ts
```diff
@@ -223,25 +243,180 @@ export class AllureReport {
...
+        testCases: JSON.parse(testCasesEntry.toString("utf8")),
+        fixtures: JSON.parse(fixturesEntry.toString("utf8")),
+        attachments: JSON.parse(attachmentsEntry.toString("utf8")),
+        environments: JSON.parse(environmentsEntry.toString("utf8")),
+        reportVariables: JSON.parse(reportVariablesEntry.toString("utf8")),
+        globalAttachments: JSON.parse(globalAttachmentsEntry.toString("utf8")),
+        globalErrors: JSON.parse(globalErrorsEntry.toString("utf8")),
+        indexAttachmentByTestResult: JSON.parse(indexAttachmentsEntry.toString("utf8")),
+      };
+      const stageTempDir = join(tmpdir(), stage);
```
delatrie: Here, `stage` is an absolute path to the stage dump. Joining it with `tmpdir()` results in an invalid path like `C:\Users\username\AppData\Local\Temp\C:\Users\username\repos\allure-framework\allure3\s3.zip`.

Either extract the stage name from this path or use `mkdtemp` as I suggested in my other comment.

packages/cli/src/commands/generate.ts
```diff
@@ -41,33 +49,53 @@ export class GenerateCommand extends Command {
...
       }
+    }
 
-      return false;
-    });
+    // don't read allure results directories without the parameter when stage file has been found
+    // or read allure results directory when it is explicitly provided
+    if (!!this.resultsDir || stageDumpFiles.length === 0) {
+      const matchedDirs = (
+        await glob(resultsDir, {
```
delatrie, line 87: Let's also use `windowsPathsNoEscape` (at least on Windows). Otherwise, it's too surprising to get `No test results directories found matching pattern...` on Windows e.g., when using tab autocompletion (which replaces all dir separators to `\`).

delatrie (changes requested):
`allure generate` fails when restoring from stages because of bad tmp dir paths (see my comments).

packages/cli/src/commands/generate.ts
epszaw: Done

packages/core/src/report.ts
epszaw: Done

packages/core/src/report.ts
epszaw: Done

delatrie (changes requested):
Retries and environments are not fully restored.

#### Retries

Run:

```shell
yarn allure run --stage s1 -- yarn workspace @allurereport/reader run test ./test/allure1.test.ts
yarn allure run --stage s2 -- yarn workspace @allurereport/reader run test ./test/allure1.test.ts
yarn allure generate --stage s?.zip
yarn allure open
```

Select any test result: no retry entries.

image

#### Environments

Run:

```shell
yarn allure run --stage s1 --environment env1 -- yarn workspace @allurereport/reader run test ./test/allure1.test.ts
yarn allure run --stage s2 --environment env2 -- yarn workspace @allurereport/reader run test ./test/allure1.test.ts
yarn allure generate --stage s?.zip
yarn allure open
```

Environments on the report level are restored. But the lists of environments on the test result pages are not:

image

packages/core/src/store/store.ts
```diff
@@ -690,6 +696,12 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
       globalAttachments: this.#globalAttachments,
       globalErrors: this.#globalErrors,
       indexAttachmentByTestResult: mapToObject(this.indexAttachmentByTestResult),
+      indexTestResultByHistoryId: mapToObject(this.indexTestResultByHistoryId),
```
delatrie: This will dump full test result objects. We already have them in `test-results.json`. Dumping four test result indices results in x5 the size, which is huge (imagine 100k test results 2kb each - that's +780 MB).

Instead of objects, dump their keys (i.e., `Record<string, string[]>` instead of `Record<string, TestResult[]>`).

packages/core/test/store/store.test.ts
```diff
@@ -2116,6 +2146,74 @@ describe("dump state", () => {
     expect(testResults).toHaveLength(1);
     expect(testResults[0].id).toBe("test-result-id");
   });
+
+  it("should handle restoreState with missing index properties gracefully", async () => {
```
delatrie, line 2150: Let's test with two dumps and validate the indices (retries, etc) are filled up correctly

packages/core/src/report.ts
```diff
@@ -223,25 +243,221 @@ export class AllureReport {
...
+          await writeFile(attachmentFilePath, attachmentContentEntry);
+
+          resultsAttachments[attachmentId] = new PathResultFile(attachmentFilePath, attachmentId);
+        }
+      } catch (err) {
+        console.error(`Can't restore state from "${stage}", continuing without it`);
+        console.error(err);
+      }
+
+      await this.#store.restoreState(dumpState, resultsAttachments);
```
delatrie, line 440: How about adding an stdout line that tells the user we've successfully restored from something-something.zip?

epszaw: Added 👍

PR #315, opened 2025-10-06
feat: add jira integration

vdvukhzhilov:


packages/cli/src/commands/jira/clear.ts
```diff
@@ -0,0 +1,126 @@
...
+  });
+
+  webhook = Option.String("--webhook", {
+    description: "Allure Jira Integration Webhook URL (default: ALLURE_JIRA_WEBHOOK environment variable)",
+    required: false,
+  });
+}
+
+export class JiraClearCommand extends BaseJiraCommand {
+  static paths = [["jira", "clear"]];
```
epszaw, line 29: If you plan to add more Jira-specific commands, let's create a `commands/jira` directory and keep all the commands separate. Then, group them using clipation command groups. 

See `results` commands as a reference.

packages/plugin-jira/test/index.test.ts
```diff
@@ -0,0 +1,3 @@
+import { it } from "vitest";
```
epszaw: Let's add some simple tests that validate that the plugin works with mocked data sources.

packages/plugin-jira/src/plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+const TRUE_VALUES = ["true", "1"];
+
+export interface JiraPluginOptions {
+  /**
+   * Url to the Allure Forge App webhook
+   * @example "https://95f453e..."
+   */
+  webhook?: string;
+  /**
+   * Generated API token
```
epszaw:
```suggestion
   * Generated Jira API token
```

packages/plugin-jira/src/plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+        throw new Error(`[${this.#pluginName}] Allure Jira Integration app error: ${error.message}`);
+      }
+      throw error;
+    }
+  }
+
+  async #uploadResults(context: PluginContext, store: AllureStore, returnData = false) {
+    this.#verifyOptions();
+
+    const statistic = await store.testsStatistic();
```
epszaw: Request `allTestResults` instead of statistics. Under the hood, the store utilizes the same data, so if you want to check that there are no test results, just check the data directly.

packages/plugin-jira/src/plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+    await this.#requestForgeApp({ operation: "upload-results", payload });
+  }
+
+  async #getReportStatus(store: AllureStore) {
+    const globalErrors = await store.allGlobalErrors();
+    const globalExitCode = await store.globalExitCode();
+    const code = globalExitCode?.actual ?? globalExitCode?.original;
+    const hasGlobalErrors = globalErrors.length > 0;
+
+    return hasGlobalErrors ? "failed" : code === 0 ? "passed" : "failed";
```
epszaw: The global exit code will work only in the case when the report is generated as a result of the `allure run` command. If there's no `globalExitCode` check that the report doesn't contain any failed or broken tests; otherwise, it should be treated as a failed one.

packages/plugin-jira/src/plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  async #getCiInfo(context: PluginContext) {
+    const jobUrl = context.ci?.pullRequestUrl ?? context.ci?.jobUrl ?? context.ci?.jobRunUrl;
+    const jobLabel = context.ci?.pullRequestName ?? context.ci?.jobName ?? context.ci?.jobRunName;
+
+    return { url: jobUrl, label: jobLabel };
+  }
+
+  async #getReportDate(store: AllureStore) {
+    const testResults = await store.allTestResults();
```
epszaw: I recommend you request all test results once and then reuse them across the methods because there can be so much data, which can consume all the machine's resources.

packages/plugin-jira/src/plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+    return testResults.reduce((acc, { stop }) => Math.max(acc, stop || 0), 0);
+  }
+
+  async #uploadReport(context: PluginContext, store: AllureStore, returnData = false) {
+    const { reportIssue } = this.#pluginOptions;
+
+    this.#verifyOptions();
```
epszaw: Actually, we can validate options once the `done` method is called because it's the entry point of the plugin and the options can't be changed in the runtime.

packages/plugin-jira/src/plugin.ts
vdvukhzhilov: I guess this is a good thing to be abstracted in some store method like `store.getReportStatus` or PluginContext field 🤔

vdvukhzhilov: Also, this is taken directly from https://github.com/allure-framework/allure3/blob/9b68c680e108ebdd9a704c2dff7c0b579437ced0/packages/web-awesome/src/components/ReportHeader/index.tsx#L31

packages/plugin-api/src/store.ts
```diff
@@ -40,6 +40,7 @@ export interface AllureStore {
   metadataByKey: <T>(key: string) => Promise<T | undefined>;
   testResultsByTcId: (tcId: string) => Promise<TestResult[]>;
   attachmentsByTrId: (trId: string) => Promise<AttachmentLink[]>;
+  retriesByTr: (tr: TestResult) => Promise<TestResult[]>;
```
epszaw, line 44: Don't see the method's implementation. Remove it from the interface. `retriesByTrId` does the same, but works with `id`.

packages/plugin-jira/src/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+      trFromMap = trMap[trId] = {
+        id: trId,
+        runs: [],
+        issue: jiraLink,
+        name: tr.name,
+        keyParams: tr.parameters.filter((p) => !p.excluded && !p.hidden),
+      };
+    }
+
+    trFromMap.runs.push({ status: tr.status, env: tr.environment, date: tr.stop! });
```
epszaw: Am I right that the `runs` field represents all attempts/retries of the test? If it does, you should build the results lists, including `hidden` test results, because retries are hidden by default.

packages/plugin-api/src/store.ts
vdvukhzhilov: https://github.com/allure-framework/allure3/blob/main/packages/core/src/store/store.ts#L519 ?

packages/plugin-jira/src/helpers.ts
vdvukhzhilov: Discussed in Slack

vdvukhzhilov:
> Can't use the plugin even it's deined in the config:
> 
> Image

Cannot reproduce

packages/sandbox/allurerc.mjs
```diff
@@ -54,6 +54,15 @@ export default defineConfig({
     ],
   },
   plugins: {
+    jira: {
```
epszaw: Remove the `jira` plugin or add `enabled: false` to disable the plugin because it doesn't work with the provided configuration:

Image

PR #321, opened 2025-10-15
Allure service public mvp

epszaw:


packages/sandbox/allurerc.mjs
```diff
@@ -54,42 +54,43 @@ export default defineConfig({
...
-      },
-    },
-    classic: {
-      options: {
-        reportName: "HelloWorld",
-        singleFile: false,
-        reportLanguage: "en",
-      },
-    },
+    // allure2: {
```
todti: leftovers?

packages/service/src/service.ts
```diff
@@ -240,12 +243,16 @@ export class AllureServiceClient {
       content = await readFile(filepath!);
     }
 
+    if (content.length > ASSET_MAX_FILE_SIZE) {
```
todti, line 246: Is variable can be customised from config in advance?

epszaw: I don't think so. We'll have the same limits on the server side, and this check is just to prevent additional requests.

PR #323, opened 2025-10-17
feat: add SFBU trend and cumulative trend charts with corresponding data accessors

Viperwow:


vdvukhzhilov:
Closing this one since the FBSU pyramid has already been implemented, and this version doesn’t  really match the requirements — it uses a trend chart instead of the intended diverging bar pyramid format.

PR #325, opened 2025-10-21
feat: add support for non mainline development tests by displaying a version label in awesome report

WolfLickefett:
Allure3 seems to be mainly development to test systems in mainline development. We want to use allure3 for our test reports, however it's important for us to display the version of the system under test for each run.

I've implemented a feature that enables the selection of a label that indicates the version number for each run.
The implementation was done with the help of Copilot Business, public code matching turned of.   
The version will be displayed in the history and retries like this:
image
At the previous result tooltip:
image
And in trend charts:
image

Please let me know if I should change something or this feature is not something you would consider for allure3.

### DIFF
--- a/packages/plugin-awesome/src/charts.ts
+++ b/packages/plugin-awesome/src/charts.ts
@@ -20,11 +20,42 @@ export const generateCharts = async (
     history: HistoryDataPoint[];
   },
 ): Promise<GeneratedChartsData | undefined> => {
-  const { charts } = options;
+  const { charts, versionLabel } = options;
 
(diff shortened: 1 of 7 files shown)


WolfLickefett:
@vdvukhzhilov is allure3 accepting external contributions at this stage or should we create a fork until the release?

baev:
Thanks for the contribution!

I’m not sure what you are trying to implement here.

If you are still interested in this feature, please create a ticket instead and explain the issue you are trying to solve, along with the setup you have.

PR #344, opened 2025-11-05
Expose Allure config type in allure package

epszaw:


packages/cli/src/index.ts
```diff
@@ -70,5 +70,5 @@ cli.runExit(args);
 
 console.log(cwd());
```
todti: Is needed? it reveals every time when using cli 😁

PR #365, opened 2025-11-21
Fix git branch detection

epszaw:


packages/service/src/history.ts
```diff
@@ -3,11 +3,18 @@ import type { AllureServiceClient } from "./service.js";
...
 
 export class AllureRemoteHistory implements AllureHistory {
-  constructor(readonly allureServiceClient: AllureServiceClient) {}
+  constructor(
+    readonly allureServiceClient: AllureServiceClient,
+    readonly branch?: string,
+  ) {}
+
+  async readHistory() {
+    if (!this.branch) {
```
delatrie, line 12: Is it possible to use Allure 3 and Allure Service outside of git context? (e.g., another vcs or no vcs at all). If so, it's possible there is no branch defined (this.branch is an empty string), yet we should still be able to save/load the history/reports.

epszaw: Let's do it in the next iteration. I think most projects utilize git and this shouldn't be a problem. Allure Service works with branches by default, so it's not possible to use the service locally without any branch.

PR #372, opened 2025-11-27
awesome: combined tree titlePath and labels

### LINKED ISSUE #262: Combined grouping by labels and titlePath
#240 introduced titlePath support by implementing `createTreeByTitlePath`. Currently, users can switch back to `createTreeByLabels`.

The next step is to combine `createTreeByLabels` with `createTreeByTitlePath` into a single function that creates a tree from a sequence of labels, optionally followed by `titlePath`.

  - The function takes a sequence of label names (empty by default).
  - The function takes a flag that appends `titlePath` at the end of each test's path (`true` by default). Alternatively, it may take the `titlePath` literal as the last label, which is treated specially.
  - The function takes a flag that preserves the nesting level in case a label's value is missing (`true` by default if `titlePath` is used; otherwise, `false`).

The combined function will cover most of the grouping cases.

todti:
fix #262 
combine `labels` and `titlePath` to create `allure-awesome` tree

### DIFF
--- a/packages/plugin-api/test/treeByLabelsAndTP.test.ts
+++ b/packages/plugin-api/test/treeByLabelsAndTP.test.ts
@@ -1,7 +1,7 @@
 import type { TestResult } from "@allurereport/core-api";
 import { randomUUID } from "node:crypto";
 import { describe, expect, it } from "vitest";
-import { createTreeByLabelsAndTitlePath } from "../src/utils/tree.js";
+import { createTreeByLabelsAndTitlePath } from "../src/index.js";
 
 const itResult = (args: Partial<TestResult> & { titlePath?: string[] }): TestResult => ({
   id: randomUUID(),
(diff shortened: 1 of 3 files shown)


packages/plugin-awesome/src/generators.ts
```diff
@@ -199,8 +200,12 @@ export const generateTree = async (
   tests: AwesomeTestResult[],
 ) => {
   const visibleTests = tests.filter((test) => !test.hidden);
+  const useCombined = labels.includes("titlePath");
```
epszaw: 1. Let's move the "special label" name to constants somewhere in commons, like `TITLE_PATH_LABEL_NAME`.
2. Let's rename it to `__titlePath` (or something similar) to exclude the possibility of overlapping with the user's labels.

packages/plugin-awesome/src/generators.ts
```diff
@@ -279,6 +309,28 @@ const buildTreeByTitlePath = (tests: AwesomeTestResult[]): TreeData<AwesomeTreeL
   };
 };
 
+const buildTreeByLabelsAndTitlePathCombined = (
+  tests: AwesomeTestResult[],
+  labels: string[],
+): TreeData<AwesomeTreeLeaf, AwesomeTreeGroup> => {
+  const hasTitlePathMarker = labels.includes("titlePath");
+  const labelNames = hasTitlePathMarker ? labels.filter((name) => name !== "titlePath") : labels;
```
epszaw: I think we can just filter them out without calling the `includes` method and then check the result array's length.

packages/plugin-awesome/src/generators.ts
```diff
@@ -199,8 +200,12 @@ export const generateTree = async (
   tests: AwesomeTestResult[],
 ) => {
   const visibleTests = tests.filter((test) => !test.hidden);
+  const useCombined = labels.includes("titlePath");
+
   const tree: TreeData<AwesomeTreeLeaf, AwesomeTreeGroup> = labels.length
-    ? buildTreeByLabels(visibleTests, labels)
+    ? useCombined
```
epszaw: Let's avoid such deep ternary operators. Better to have a single switch/case expression or multiple if.

packages/plugin-api/test/treeByLabelsAndTP.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    expect(names).toEqual(["Admin", "User"]);
+
+    for (const groupId of result.root.groups ?? []) {
+      const group = result.groupsById[groupId];
+      expect(group.leaves).toHaveLength(1);
+    }
+  });
+
+  it('should treat "titlePath" as special last label marker', () => {
```
epszaw: What'd happen when `titlePath` is not in the end?
If we want to force users to use `titlePath` only in this certain way, we need to validate it (at least for now).
Or if we're going to go the way of maximum freedom, we need to support `titlePath` in the middle in the future.

epszaw (changes requested):

packages/plugin-api/test/treeByLabelsAndTP.test.ts
todti: it will be just ignored

packages/plugin-awesome/src/generators.ts
```diff
@@ -197,11 +198,18 @@ export const generateTree = async (
   treeFilename: string,
   labels: string[],
   tests: AwesomeTestResult[],
+  appendTitlePath?: boolean,
```
epszaw: Let's add an optional options parameter with the `appendTitlePath` property. It's better to avoid single boolean parameters and group them into objects:

```
options?: {
  appendTitlePath?: boolean
}
```

packages/plugin-api/test/treeByLabelsAndTP.test.ts
epszaw: It seems the test is outdated. Update the test to make it check the actual logic

PR #376, opened 2025-11-30
feat: update current status pie

vdvukhzhilov:
https://github.com/user-attachments/assets/bbd0c0d7-328d-4258-af24-1cd25e38b054


packages/web-commons/src/charts/generators.ts
```diff
@@ -42,8 +42,8 @@ const generateChartData = async (props: {
       case ChartType.Trend:
         result[chartId] = generateTrendChart(chartOption, storeData, reportName)!;
         break;
-      case ChartType.Pie:
-        result[chartId] = generatePieChart(chartOption, storeData)!;
+      case ChartType.CurrentStatus:
```
epszaw, line 45: Ideally if user will have choice:
- use pre-defined chart, like current status pie chart
- use his-own pie chart configuration
At this moment, it's ok to have pre-defined charts only, but in future we should provide a fully customizable dashboard experience

packages/web-components/src/components/Charts/CurrentStatusChartWidget/parts.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  if (metricType === "percent") {
+    return <PercentsMetric layer={layerProps} metric={metric} i18n={i18n} total={total} />;
+  }
+
+  return null;
+};
+
+const TotalMetric = ({
```
epszaw, line 33: Let's export all the components. It can be useful for tests

epszaw (approved):

PR #391, opened 2025-12-10
Add summary meta

epszaw:
remove with results links property from the summary and provide it to the meta
add report id property to the summary


rakhee-puria:
[epszaw](https://github.com/epszaw) can allure-maven version be also released by getting rid of tika-core 2.9.4 ? it is blocking our JFROG scans thereby failing our builds

PR #404, opened 2025-12-17
Exit with code 1 when target results dir doesn't exist

epszaw:


packages/cli/src/commands/allure2.ts
```diff
@@ -56,6 +58,12 @@ export class Allure2Command extends Command {
   });
 
   async execute() {
+    if (!existsSync(this.resultsDir)) {
+      console.error(red(`Given test results directory doesn't exist: ${this.resultsDir}`));
```
delatrie: "Given" sounds kinda like "Let's suppose" in such a context. Alternative: `The given...`
```suggestion
      console.error(red(`The test results directory doesn't exist: ${this.resultsDir}`));
```

PR #407, opened 2025-12-17
Enchanced open command

epszaw:
The PR's summary:

- add `serve` alias for `open` command
- `open` command now generates the report and the serve it
- `run` command now opens the generated report after tests process end when `open` flag is provided
- add `open` and `port` options to the config which affect `run` and `open` commands

### DIFF
--- a/packages/cli/test/commands/open.test.ts
+++ b/packages/cli/test/commands/open.test.ts
@@ -2,7 +2,6 @@ import { readConfig } from "@allurereport/core";
 import { serve } from "@allurereport/static-server";
 import { run } from "clipanion";
 import { glob } from "glob";
-import { randomUUID } from "node:crypto";
 import { existsSync } from "node:fs";
 import { mkdtemp } from "node:fs/promises";
 import { tmpdir } from "node:os";
@@ -11,9 +10,6 @@ import { type Mock, beforeEach, describe, expect, it, vi } from "vitest";
 import { generate } from "../../src/commands/commons/generate.js";
 import { OpenCommand } from "../../src/commands/open.js";
 
-vi.mock("node:crypto", () => ({
-  randomUUID: vi.fn(),
-}));
 vi.mock("node:fs", async (importOriginal) => {
   return {
     ...(await importOriginal()),
@@ -57,12 +53,10 @@ describe("open command", () => {
   it("should generate report in temp directory and serve when no summary files found", async () => {
     const fixtures = {
       resultsDir: "allure-report",
-      uuid: "0123",
-      tmpDir: "/tmp/allure-report-0123",
+      tmpDir: "/tmp/allure-report-abc123",
     };
 
     (existsSync as Mock).mockReturnValue(true);
-    (randomUUID as Mock).mockReturnValue(fixtures.uuid);
     (tmpdir as Mock).mockReturnValue("/tmp");
     (mkdtemp as Mock).mockResolvedValue(fixtures.tmpDir);
     (readConfig as Mock).mockResolvedValue({ output: fixtures.tmpDir });
@@ -77,7 +71,7 @@ describe("open command", () => {
 
     await command.execute();
 
-    expect(mkdtemp).toHaveBeenCalledWith(join("/tmp", `allure-report-${fixtures.uuid}`));
+    expect(mkdtemp).toHaveBeenCalledWith(join("/tmp", "allure-report-"));
     expect(readConfig).toHaveBeenCalledWith(".", expect.any(Object), {
       port: expect.any(Object),
       output: fixtures.tmpDir,
@@ -139,12 +133,10 @@ describe("open command", () => {
   it("should generate report when target directory does not exist", async () => {
     const fixtures = {
       resultsDir: "allure-report",
-      uuid: "0123",
-      tmpDir: "/tmp/allure-report-0123",
+      tmpDir: "/tmp/allure-report-abc123",
     };
 
     (existsSync as Mock).mockReturnValue(false);
-    (randomUUID as Mock).mockReturnValue(fixtures.uuid);
     (tmpdir as Mock).mockReturnValue("/tmp");
     (mkdtemp as Mock).mockResolvedValue(fixtures.tmpDir);
     (readConfig as Mock).mockResolvedValue({ output: fixtures.tmpDir });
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/open.ts
```diff
@@ -24,21 +36,37 @@ export class OpenCommand extends Command {
...
 
   async execute() {
-    const config = await readConfig(this.cwd, this.config, { output: this.reportDir });
+    const cwd = this.cwd ?? processCwd();
+    const config = await readConfig(cwd, this.config, {
+      output: this.output,
+      port: this.port,
+    });
+    // if any summary.json file present in the output directory, we assume that report is already generated
+    const summaryFiles = await glob(join(config.output, "**", "summary.json"), {
```
delatrie: Using output for this is counterintuitive:

1. Differs from Allure 2: `allure open ./my-report-path` should open, not generate.
2. `npx allure open "./**/results" --output ./report` should always generate even if ./report exists.

Let's branch on `resultsDir` instead: if it's a single path to an existing report directory and no `--output` given, then open it. Otherwise, generate the report first.
Perhaps, rename `resultsDir` to `reportOrResultsDir` or smth else.

packages/cli/src/commands/open.ts
```diff
@@ -1,20 +1,32 @@
...
   });
 
-  reportDir = Option.String({ required: false, name: "The directory with Allure results" });
+  resultsDir = Option.String({
+    required: false,
+    name: "Pattern to match test results directories in the current working directory (default: ./**/allure-results)",
+  });
+
+  output = Option.String("--output,-o", {
+    description: "The output directory name. Absolute paths are accepted as well (default: allure-report)",
```
delatrie: We should generate into a temp directory by default:

  - this way, we won't overwrite an existing report
  - it better suits the idea of `allure serve` in Allure 2, which is "I don't care about the physical report, just open these allure results quickly`

packages/cli/src/commands/open.ts
```diff
@@ -24,21 +34,48 @@ export class OpenCommand extends Command {
...
   cwd = Option.String("--cwd", {
     description: "The working directory for the command to run (default: current working directory)",
   });
 
   async execute() {
-    const config = await readConfig(this.cwd, this.config, { output: this.reportDir });
+    const cwd = this.cwd ?? processCwd();
+
+    const isGlobPatternGiven = this.resultsDir.includes("*") || this.resultsDir.includes("?");
+    const summaryFilesGlob = isGlobPatternGiven ? this.resultsDir : join(this.resultsDir, "**", "summary.json");
```
delatrie: The `servePath` option of `serve` doesn't support globs, so there is no sense in looking for summary files if a glob is given to `allure open`.

Let's instead check if `resultsDIr` points to an existing directory, and only in this case, check `summary.json`:

```js
const maybeReportDir = resolve(cwd, servePath);
let directoryExists = false;
try {
  directoryExists = (await stat(maybeReportDir)).isDirectory();
} catch {}

const hasSummaryFile = directoryExists
  ? (await glob(
    join(maybeReportDir, "**", "summary.json"),
    {
      // glob options...
    }
  ))
  : false;

if (!hasSummaryFile) {
  // generate
}
```

If `resultsDir` is given and doesn't directly point to an existing directory (e.g., because it's a glob pattern), `generate` will handle this.

packages/cli/src/commands/open.ts
```diff
@@ -1,20 +1,30 @@
...
-      ["open", "Serve the default report directory"],
-      ["open custom-report --port 8080 --live", "Serve the custom-report directory on port 8080 with live reload"],
+      ["open ./allure-results", "Generate and serve the report based on given test results directory"],
+      ["open --port 8080 ./allure-report", "Serve the report on port 8080"],
     ],
   });
 
-  reportDir = Option.String({ required: false, name: "The directory with Allure results" });
+  resultsDir = Option.String({
+    required: true,
```
delatrie: `allure open` with no arguments should serve `allure-report` by default (just like it currently does)

packages/cli/src/commands/open.ts
```diff
@@ -24,21 +34,48 @@ export class OpenCommand extends Command {
...
+      windowsPathsNoEscape: true,
+      cwd,
+    });
+    let config: FullConfig;
+
+    // there's no generated report to serve, so we generate it first in a temp directory
+    if (!summaryFiles.length) {
+      config = await readConfig(cwd, this.config, {
+        port: this.port,
+        output: join(cwd, ".allure", randomUUID()),
```
delatrie: Let's use mkdtemp + temp and remove the directory after the serving is done by ctrl+c (reports may easily weigh a hundred MB; 10 opens and we've eaten a gig...)

packages/cli/src/commands/open.ts
```diff
@@ -1,20 +1,31 @@
...
-      ["open", "Serve the default report directory"],
-      ["open custom-report --port 8080 --live", "Serve the custom-report directory on port 8080 with live reload"],
+      ["open ./allure-results", "Generate and serve the report based on given test results directory"],
+      ["open --port 8080 ./allure-report", "Serve the report on port 8080"],
     ],
   });
 
-  reportDir = Option.String({ required: false, name: "The directory with Allure results" });
+  resultsDir = Option.String({
+    name: "Pattern to match test results directories in the current working directory (default: ./allure-results)",
```
delatrie:
```suggestion
    name: "A report to open or a pattern to match test results directories in the current working directory (default: ./allure-results)",
```

packages/cli/src/commands/open.ts
```diff
@@ -24,22 +35,59 @@ export class OpenCommand extends Command {
...
+        port: this.port,
+      });
+
+      await serve({
+        port: config.port ? parseInt(config.port, 10) : undefined,
+        servePath: targetFullPath,
+        open: true,
+      });
+    } else {
+      const tmpDir = await mkdtemp(join(tmpdir(), `allure-report-${randomUUID()}`));
```
delatrie: `mkdtemp` returns a random path already

```suggestion
      const tmpDir = await mkdtemp(join(tmpdir(), `allure-report-`));
```

packages/cli/src/commands/open.ts
```diff
@@ -24,22 +35,59 @@ export class OpenCommand extends Command {
...
+
+      await generate({
+        resultsDir: targetFullPath,
+        cwd,
+        config,
+      });
+
+      // clean up temp report directory on ctrl-c
+      process.on("SIGINT", async () => {
+        await rm(config.output, { recursive: true });
```
delatrie: Better wrap it in try-catch to prevent the app from hanging if `rm` fails. Otherwise, `process.exit` won't be called.

PR #429, opened 2025-12-29
Add history limit

### LINKED ISSUE #417: Allow configuring report history size via CLI flag
**Describe the feature**
Introduce a CLI flag that allows users to configure how many historical entries are preserved when generating a new report.

**Motivation**
Currently, the number of entries kept in the history file is unlimited. Providing a configurable option would give users better control over report size and historical depth.

**Proposed usage**
```bash
pnpm exec allure generate --history=20
```
Keep only 20 lines.

epszaw:
fixes #417

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -76,29 +76,51 @@ export const writeHistory = async (historyPath: string, data: HistoryDataPoint) 
 };
 
 export class AllureLocalHistory implements AllureHistory {
-  constructor(private readonly historyPath: string) {}
+  constructor(
+    private readonly params: {
+      historyPath: string;
+      limit?: number;
+    },
+  ) {}
 
   async readHistory() {
-    const path = resolve(this.historyPath);
+    const path = resolve(this.params.historyPath);
 
     try {
-      return (await readFile(path, { encoding: "utf-8" }))
+      const historyPoints = (await readFile(path, { encoding: "utf-8" }))
         .split("\n")
         .filter((line) => line && line.trim() !== "")
         .map((line) => JSON.parse(line) as HistoryDataPoint);
+
+      if (this.params.limit) {
+        return historyPoints.slice(-this.params.limit);
+      }
+
+      return historyPoints;
     } catch (e) {
       if (isFileNotFoundError(e)) {
         return [];
       }
+
       throw e;
     }
   }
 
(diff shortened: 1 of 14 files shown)


packages/core/src/history.ts
```diff
@@ -76,29 +76,51 @@ export const writeHistory = async (historyPath: string, data: HistoryDataPoint)
...
+    },
+  ) {}
 
   async readHistory() {
-    const path = resolve(this.historyPath);
+    const path = resolve(this.params.historyPath);
 
     try {
-      return (await readFile(path, { encoding: "utf-8" }))
+      const historyPoints = (await readFile(path, { encoding: "utf-8" }))
```
delatrie, line 90: As soon as the history file hits 512 MiB in size (which seems realistic, given each history point stores the entire test results), `readFile` will throw `Invalid string length`. We should use `node:readline`. There is an example of line-by-line reading in the docs: https://nodejs.org/api/readline.html#example-read-file-stream-line-by-line. Since we're using JSONL, applying the limit should be straightforward.

packages/core/src/history.ts
```diff
@@ -76,29 +76,51 @@ export const writeHistory = async (historyPath: string, data: HistoryDataPoint)
...
-    await writeFile(path, `${JSON.stringify(data)}\n`, { encoding: "utf-8", flag: "a+" });
+
+    if (!this.params.limit) {
+      await writeFile(path, `${JSON.stringify(data)}\n`, { encoding: "utf-8", flag: "a+" });
+      return;
+    }
+
+    const existingHistory = await this.readHistory();
+    const updatedHistory = [...existingHistory, data].slice(-this.params.limit);
+    const fileContent = updatedHistory.reduce((acc, point) => `${acc}${JSON.stringify(point)}\n`, "");
```
delatrie, line 122: The same concerns about hitting the max string length apply here. Instead of concatenating a single giant string, we should do the following:

  - read line-by-line and store the lines in an array separately
  - apply the limit
  - write the lines to the file
  - append the new entry

delatrie (approved):

packages/core/src/history.ts
epszaw: This problem can appear for users who already have a long history. I think, there is not so much of them. New users won't face with the problem if they limit the history because the file can't contain more lines than the limit

PR #434, opened 2026-01-07
New allure service access tokens

epszaw:
introduct new access tokens logic to the service


packages/service/src/service.ts
```diff
@@ -2,96 +2,70 @@ import { type HistoryDataPoint } from "@allurereport/core-api";
...
+  decodeToken(token: string) {
+    try {
+      const base64Url = token.split(".")[1];
+      const base64 = base64Url.replace(/-/g, "+").replace(/_/g, "/");
+      const jsonPayload = decodeURIComponent(
+        atob(base64)
+          .split("")
+          .map((c) => `%${`00${c.charCodeAt(0).toString(16)}`.slice(-2)}`)
+          .join(""),
+      );
```
delatrie: This is basically a noop :)

```suggestion
      const jsonPayload = atob(base64);
```

packages/service/src/service.ts
```diff
@@ -134,15 +108,10 @@ export class AllureServiceClient {
...
-
-    if (!this.currentProjectUuid) {
-      throw new Error("Project is not set");
-    }
-
     const { history } = await this.#client.get<{ history: HistoryDataPoint[] }>(
       limit
-        ? `/projects/${this.currentProjectUuid}/${encodeURIComponent(branch)}/history?limit=${limit}`
-        : `/projects/${this.currentProjectUuid}/${encodeURIComponent(branch)}/history`,
+        ? `/projects/history/${encodeURIComponent(branch)}?limit=${limit}`
```
delatrie: We don't validate `limit` to be a number at runtime. Better encode it too:
```suggestion
        ? `/projects/history/${encodeURIComponent(branch)}?limit=${encodeURIComponent(limit)}`
```

PR #451, opened 2026-01-14
feat: add aql parser

vdvukhzhilov:
Closes http://github.com/allure-framework/allure-org/issues/265


package.json
```diff
@@ -76,7 +80,8 @@
     "prettier-plugin-packagejson": "^2.5.2",
     "rimraf": "^6.0.1",
     "tslib": "^2.7.0",
-    "typescript": "^5.6.3"
+    "typescript": "^5.6.3",
+    "vitest": "^2"
```
epszaw: Don't add `vitest` and `allure-vitest` to the root. Install it in the target package only

package.json
```diff
@@ -35,6 +35,8 @@
   "resolutions": {
     "@preact/compat": "^18.3.1",
     "@preact/signals": "^2.5.1",
+    "allure-vitest": "^3.4.4",
+    "allure-js-commons": "^3.4.4",
```
epszaw: Why do we need to set resolutions for these dependencies here? Update the dependencies directly in the packages using `yarn upgrade-interactive` command

package.json
```diff
@@ -52,6 +54,7 @@
     "semver": "^7.5.2",
     "tar": "^6.2.1",
     "vite": "^5.4.17",
+    "vitest": "^2",
```
epszaw: As far as I remember, we support latest versions of the framework, so no need to force v2 there

packages/aql/src/model.ts
```diff
@@ -0,0 +1,203 @@
+/**
```
epszaw, line 1: Usually, we call files with the package's types `model.ts`

packages/aql/src/parser/index.ts
```diff
@@ -0,0 +1,511 @@
...
+    if (this.isEOF()) {
+      return { expression: null };
+    }
+
+    const expression = this.parseExpression();
+    this.expect("EOF");
+    return { expression };
+  }
+
+  private parseExpression(): AqlExpression {
```
epszaw: There's no need for aliases, keep the original methods only

epszaw: Furthermore, the method actually parses `or` expression instead of parsing an entire aql one

packages/aql/src/utils/index.ts
```diff
@@ -0,0 +1,143 @@
...
+ * ```typescript
+ * const result = validateAql('status = "passed"');
+ * if (result.valid) {
+ *   console.log("AQL is valid");
+ * } else {
+ *   console.error("Error:", result.error);
+ * }
+ * ```
+ */
+export function validateAql(aql: string, config?: AqlParserConfig): ValidateAqlResult {
```
epszaw: I think we don't need the utility, but need just trying to parse a query.
In case if we validate the expression first and then parse it one more time – we do extra job, so the utility looks a bit useless

packages/aql/src/tokenizer/index.ts
```diff
@@ -0,0 +1,385 @@
...
+    this.tokens = [];
+
+    while (this.position < this.input.length) {
+      const token = this.nextToken();
+      if (token.type !== "WS") {
+        this.tokens.push(token);
+      }
+    }
+
+    this.tokens.push({ type: "EOF", value: "", position: this.position });
```
epszaw: I'm not sure, that we need `EOF` token because mostly aql is present as a single line

epszaw: But in case when `EOF` is used to explicitly mark the end of the tokens sequence – we can leave it

packages/aql/src/tokenizer/index.ts
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    // Skip whitespace
+    if (this.isWhitespace(this.currentChar())) {
+      return this.readWhitespace();
+    }
+
+    // Operators (multi-character first)
+    if (this.match(">=")) {
+      return { type: "GE", value: ">=", position: startPos };
```
epszaw: I guess, current `value` doesn't reflect well it's purpose, better to rename it to `source` or `raw` because in my vision, `value` property can contain operands of the token, but not the raw expression itself. Let's rename the property

packages/aql/src/tokenizer/index.ts
```diff
@@ -0,0 +1,385 @@
...
+      return { type: "RBRACKET", value: "]", position: startPos };
+    }
+    if (this.currentChar() === ",") {
+      this.advance();
+      return { type: "COMMA", value: ",", position: startPos };
+    }
+
+    // Strings
+    if (this.currentChar() === '"') {
+      return this.readString();
```
epszaw, line 110: Check, that the method correctly handles another possible tokens, such as brackets and parentheses

packages/aql/src/tokenizer/index.ts
```diff
@@ -0,0 +1,385 @@
+/**
+ * Tokenizer for AQL
+ */
+import { AqlErrors } from "../errors/index.js";
+
+export type TokenType =
```
epszaw: Move to the file with types

packages/aql/src/tokenizer/index.ts
```diff
@@ -0,0 +1,385 @@
...
+  | "FUNCTION"
+  | "LPAREN"
+  | "RPAREN"
+  | "LBRACKET"
+  | "RBRACKET"
+  | "COMMA"
+  | "EOF"
+  | "WS";
+
+export interface Token {
```
epszaw: Same here. Let's keep all package's types in a single file

packages/aql/src/tokenizer/index.ts
```diff
@@ -0,0 +1,385 @@
...
+  type: TokenType;
+  value: string;
+  position: number;
+}
+
+// Character class checks - using character codes for better performance
```
epszaw: Move the implementation details to the class' jsdoc block

epszaw (changes requested):

packages/aql/src/tokenizer/index.ts
vdvukhzhilov: > EOF is used to explicitly mark the end of the tokens sequence

Yes, this is the case

packages/aql/src/tokenizer/index.ts
vdvukhzhilov: Yeap, agreed. I left `value` only where it is needed - like identifiers or identifiers's values

packages/aql/src/tokenizer/index.ts
vdvukhzhilov: `readString` reads all symbols until ending `"`, including `()[]` and other special characters

packages/aql/src/errors/index.ts
```diff
@@ -0,0 +1,260 @@
...
+     */
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, this.constructor);
+    }
+  }
+
+  /**
+   * Gets a user-friendly error message
+   */
+  get userMessage(): string {
```
epszaw: Why can't we just directly read the message property?

packages/aql/src/utils/index.ts
```diff
@@ -0,0 +1,171 @@
...
+      return "null";
+    default:
+      return value.value;
+  }
+}
+
+/**
+ * Type helper for AQL expression
+ */
+export function aqlExpression(expression: AqlExpression): AqlExpression {
```
epszaw, line 127: Why can't we just use the type directly? For example, `expression as AqlExpression`?

vdvukhzhilov: It's just easier sometimes to use type helpers. E.g. when typescript is not used in project, but IDE can highlight errors still (like VS Code or WebStorm)

epszaw (approved):

PR #453, opened 2026-01-15
Central data storage for all plugins

ve-ev:
- added mechanism for creation a central data storage to prevent file duplication
- fixed issue with generation more than 20 plugins on web summary page
- fixed override plugin report name with a fallback


baev:
@ve-ev Thanks for the contribution!

Unfortunately, we will not accept this contribution at this time. Allure3 is still under active development, so it is not yet the right time for these optimizations.

We will provide these optimizations ourselves once the core and plugin API have stabilized.

PR #454, opened 2026-01-15
Refactor report filtering, add tags filter

vdvukhzhilov:
Closes https://github.com/allure-framework/allure-org/issues/260
Closes https://github.com/allure-framework/allure3/issues/104

## Summary of Changes

- Refactored routing to be fully URL-driven:
  - Routing is now based on hash and search parameters.
  - All navigation goes through the history API and is fully reactive.
  - Application state is synchronized with the URL and can be restored from it.
  - Search params are kept between navigations so that filters state is persistent

- Introduced tag-based filtering:
  - Test results with tags in metadata now display clickable tags.
  - Clicking a tag opens a new window with the corresponding tag filter applied.
  - Tag filter links can also be copied and shared.

- Moved filters persistence from Local Storage to URL:
  - All filters are now stored in search parameters.
  - This allows sharing filtered views via URL.
  - Only sorting preferences (field and order) remain in Local Storage, as they are considered user-specific.

- Reworked filtering engine:
  - All filters are translated into AQL and compiled into an AST.
  - Filtering is executed based on this generated query.

- Status bar is now part of the filtering system:
  - Selecting status tabs (e.g. Failed, Broken) updates the URL.
  - These filters are handled through the same AQL mechanism.

- Added extensible filter architecture:
  - The system now supports adding more complex filters in the future (e.g. via a dedicated filter form).

- Transitions and Retry/Flaky moved into dedicated filter panels:
  - Transition filters (New, Regressed, No Function) are applied using OR logic.
  - Retry and Flaky filters also use OR logic when combined.
  - Tag filters use OR logic as well.

- Improved tag handling:
  - The tag list is aggregated across all environments and trees.
  - A maximum of 20 tags can be applied simultaneously (mainly for performance reasons).
  - If more than 20 tags are selected, only 20 are really used.
  - The UI highlights this state and shows a tooltip explaining the limitation.

### DIFF
--- a/packages/web-awesome/src/components/Header/index.tsx
+++ b/packages/web-awesome/src/components/Header/index.tsx
@@ -5,7 +5,6 @@ import { HeaderControls } from "@/components/HeaderControls";
 import { SectionPicker } from "@/components/SectionPicker";
 import { TrBreadcrumbs } from "@/components/TestResult/TrHeader/TrBreadcrumbs";
 import { testResultRoute } from "@/stores/router";
-import { availableSections } from "@/stores/sections";
 import { currentTrId } from "@/stores/testResult";
 import { testResultStore } from "@/stores/testResults";
 import { CiInfo } from "./CiInfo";
@@ -21,7 +20,7 @@ const testResult = computed(() => testResultStore.value?.data?.[currentTrId.valu
 export const Header = ({ className }: HeaderProps) => {
   return (
     <div className={clsx(styles.above, className)}>
-      {Boolean(availableSections.value?.length) && <SectionPicker />}
+      <SectionPicker />
       {!isTestResultRoute.value && <CiInfo />}
       {isTestResultRoute.value && <TrBreadcrumbs testResult={testResult.value} />}
       <HeaderControls className={styles.right} />
--- a/packages/web-awesome/src/components/SectionPicker/index.tsx
+++ b/packages/web-awesome/src/components/SectionPicker/index.tsx
@@ -37,7 +37,7 @@ export const SectionPicker = () => {
       )}
     >
       <Menu.Section>
-        {["default", ...availableSections.value].map((value) => (
+        {["default", ...availableSections].map((value) => (
           <Menu.ItemWithCheckmark
             onClick={() => setSection(value)}
             key={value}
--- a/packages/web-awesome/src/index.tsx
+++ b/packages/web-awesome/src/index.tsx
@@ -15,7 +15,7 @@ import { fetchPieChartData } from "@/stores/chart";
 import { currentEnvironment, environmentsStore, fetchEnvironments } from "@/stores/env";
 import { fetchEnvInfo } from "@/stores/envInfo";
 import { fetchGlobals } from "@/stores/globals";
-import { getLayout, isLayoutLoading, layoutStore } from "@/stores/layout";
+import { isLayoutLoading, layoutStore } from "@/stores/layout";
 import { fetchTestResult, fetchTestResultNav } from "@/stores/testResults";
 import { fetchEnvTreesData } from "@/stores/tree";
 import { isMac } from "@/utils/isMac";
@@ -54,7 +54,7 @@ const App = () => {
     ];
 
     if (globalThis) {
-      fns.unshift(getLocale, getLayout as () => Promise<void>, getTheme as () => Promise<void>);
+      fns.unshift(getLocale, getTheme as () => Promise<void>);
     }
 
     await waitForI18next;
--- a/packages/web-awesome/src/stores/layout.ts
+++ b/packages/web-awesome/src/stores/layout.ts
@@ -2,7 +2,26 @@ import { getReportOptions } from "@allurereport/web-commons";
 import { computed, signal } from "@preact/signals";
 import type { AwesomeReportOptions, Layout } from "types";
 
-export const layoutStore = signal<Layout>("base");
+const reportOptions = getReportOptions<AwesomeReportOptions>();
+
+const DEFAULT_LAYOUT = "base";
+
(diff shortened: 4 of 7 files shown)


packages/e2e/test/allure-awesome/features/flaky.test.ts
```diff
@@ -131,17 +131,6 @@ test.describe("flaky", () => {
     expect(flaky).toBe("1");
   });
 
-  test("should show tooltip with flaky filter description on hover", async ({ page }) => {
```
vdvukhzhilov, line 134: I’ve removed the tooltip tests because UI details like tooltips should be covered by other layers of tests, not end-to-end ones

packages/web-components/src/components/SearchBox/index.tsx
```diff
@@ -34,22 +40,44 @@ export const SearchBox = (props: Props) => {
...
+        aria-invalid={hasError || undefined}
       />
+      {trailingSlot && (
+        <div className={styles.slot} onClick={stopPropagation}>
+          {trailingSlot}
+        </div>
+      )}
       {showClear && (
-        <div className={styles.clearButton}>
+        <div className={styles.clearButton} onClick={stopPropagation}>
```
epszaw: Doesn't `IconButton` handler already prevent propagation? Is it certain to add additional handler on top of it?

packages/web-commons/vitest.config.ts
```diff
@@ -1,11 +1,18 @@
...
 export default defineConfig({
+  resolve: {
+    alias: {
+      "@": resolve(__dirname, "./src"),
+    },
+  },
   test: {
-    include: ["./test/**/*.test.ts"],
+    environment: "jsdom",
+    include: ["./test/**/*.test.ts", "./src/**/*.test.ts"],
```
epszaw: Maybe it's better just to move all tests from `src` to `test` directory

packages/web-commons/src/utils.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Creates a debounced function that delays invoking func until after wait milliseconds
+ * have elapsed since the last time the debounced function was invoked.
+ * @param func The function to debounce.
+ * @param wait The number of milliseconds to delay.
+ * @returns Returns the new debounced function.
+ */
+export const debounce = <T extends (...args: any[]) => any>(
```
epszaw: Let's just use ready implementation, like [debounce](https://www.npmjs.com/package/debounce)

packages/web-commons/test/stores/url/store.test.ts
```diff
@@ -0,0 +1,103 @@
+import { label } from "allure-js-commons";
```
epszaw, line 1: Move the test to `test` directory like in the rest packages

packages/web-awesome/src/components/SplitLayout/index.tsx
```diff
@@ -28,16 +30,20 @@ const Loader = () => {
...
 export const SplitLayout = () => {
-  const testResultId = route.value.params?.testResultId ?? null;
+  const testResultId = currentTrId.value;
   const [cachedMain, setCachedMain] = useState<JSX.Element | null>(null);
   const { t } = useI18n("controls");
   const leftSide = (
     <Loadable source={treeStore} renderLoader={() => <PageLoader />} renderData={() => <MainReportWrapper />} />
   );
 
+  console.log(isTestResultRoute.value);
```
epszaw: Leftover

packages/web-awesome/src/components/Tree/index.tsx
```diff
@@ -4,19 +4,24 @@ import { MetadataButton } from "@/components/MetadataButton";
 import { reportStatsStore, statsByEnvStore } from "@/stores";
 import { collapsedEnvironments, currentEnvironment, environmentsStore } from "@/stores/env";
 import { useI18n } from "@/stores/locale";
-import { navigateTo, route } from "@/stores/router";
+import { clearAwesomeFilters, treeAwesomeStatus } from "@/stores/reportFilters/store";
```
epszaw: Let's call them just filters and trees because there's no any other similary entities in context of this package

packages/web-commons/src/stores/url/store.ts
```diff
@@ -0,0 +1,91 @@
...
+    host: urlSignal.value.host,
+    hostname: urlSignal.value.hostname,
+    href: urlSignal.value.href,
+    origin: urlSignal.value.origin,
+    pathname: urlSignal.value.pathname,
+    port: urlSignal.value.port,
+    protocol: urlSignal.value.protocol,
+    search: urlSignal.value.search,
+    searchParams: new URLSearchParams(urlSignal.value.searchParams),
+    params: Array.from(urlSignal.value.searchParams.entries()).reduce(
```
epszaw: Why can't we return the original `URL` object, but with parsed params object instead?

packages/web-awesome/src/components/SplitLayout/index.tsx
vdvukhzhilov: 😢 

Strange that lint is not failing here 🤷

packages/web-components/src/components/SearchBox/index.tsx
vdvukhzhilov: yeap, you are right

packages/web-commons/src/stores/url/store.ts
vdvukhzhilov: Well, it is not that simple.

But I'll pick only what is needed 👌

vdvukhzhilov:
> 🔴 When there's no tags render disabled dropdown
> 
> image

As the options are generated beforehand, now when there are no tags in tests - the tag filter is not even showed

PR #469, opened 2026-01-28
feat: add image diff for attachments

vdvukhzhilov:
https://github.com/user-attachments/assets/d5d40eab-4259-4c33-a56f-9681ab1ce9d8


packages/web-components/src/components/Attachment/model.ts
```diff
@@ -0,0 +1,14 @@
+import type { AttachmentTestStepResult } from "@allurereport/core-api";
+import type { AttachmentData } from "@allurereport/web-commons";
+import type { ComponentProps } from "preact";
+import type { ImageDiff } from "../ImageDiff";
+
+export type AttachmentProps = {
+    attachment: AttachmentData | null;
```
epszaw: Something happened with formatting. Run format script

packages/web-components/src/components/Attachment/AttachmentImage.tsx
```diff
@@ -1,27 +1,28 @@
...
-  attachment: { img: string; originalFileName: string };
-}> = ({ attachment }) => {
+export const AttachmentImage = (props: AttachmentProps) => {
+  const { attachment, item } = props;
+
   const [isValidImage, setIsValidImage] = useState(true);
 
   useEffect(() => {
-    if (attachment?.img) {
+    if (attachment && "img" in attachment && attachment.img) {
```
epszaw, line 11: Why it's not enough just to check `attachment?.img`?

vdvukhzhilov: Nope, because it may have `img` or may not, depending on atachment type. So to keep TS happy this check is needed

epszaw (approved):

Issue #442, opened 2026-01-12
Add English (ISO-8601) locale


epszaw:
Relates to allure-framework/allure2#1195

PR #482, opened 2026-02-03
ISO-8601 locale for date and time formatting

todti:
#442 
- Added `en-iso` (English ISO‑8601) as a locale with configurable date/time formatting via `LOCALE_DATE_TIME_OVERRIDES`.
- Fixed Ukrainian locale code from `ua` to `uk` and renamed locale files.
- `en-iso` now loads en JSON resources.
-


packages/web-classic/src/utils/time.ts
```diff
@@ -12,6 +13,19 @@ const defaultOptions: Intl.DateTimeFormatOptions = {
...
+    ...options,
+    ...(override?.options ?? {}),
+  });
+  const formatted = formatter.format(date);
+
+  if (override?.includeAtSeparator === false || override?.stripComma) {
+    return formatted.replace(",", "");
+  }
+
+  return formatted.replace(",", ` ${t("at")}`);
```
epszaw:
```suggestion
  return formatted.replace(",", t("at"));
```

PR #485, opened 2026-02-06
Global attachments and errors

epszaw:


packages/core/src/store/store.ts
```diff
@@ -133,7 +135,7 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
   readonly indexFixturesByTestResult: Map<string, TestFixtureResult[]> = new Map<string, TestFixtureResult[]>();
   readonly indexKnownByHistoryId: Map<string, KnownTestFailure[]> = new Map<string, KnownTestFailure[]>();
 
-  #globalAttachments: AttachmentLink[] = [];
+  #globalAttachmentsIds: string[] = [];
```
delatrie: nit(grammar): noun adjunct "attachment" must be in singular
```suggestion
  #globalAttachmentIds: string[] = [];
```

packages/core/src/store/store.ts
```diff
@@ -375,7 +390,33 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
-    Object.keys(metadata).forEach((key) => this.#metadata.set(key, metadata[key]));
+    Object.keys(metadata).forEach((key) => {
+      this.#metadata.set(key, metadata[key]);
+    });
+  }
+
+  async visitGlobals(globals: RawGlobals): Promise<void> {
+    const { errors, attachments } = globals;
+
+    errors.forEach((error) => {
```
delatrie: nit: alternative:

```js
this.#globalErrors.push(...errors);
```

packages/core/src/store/store.ts
```diff
@@ -262,8 +270,16 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
-    return this.#globalAttachments;
+  async allGlobalAttachments(): Promise<AttachmentLinkLinked[]> {
+    return this.#globalAttachmentsIds.reduce((acc, id) => {
+      const attachment = this.#attachments.get(id);
+
+      if (!attachment) {
+        return acc;
+      }
+
+      return acc.concat(attachment as AttachmentLinkLinked);
```
delatrie: `reduce` + `concat` creates a quadratic complexity out of nowhere. Better mutate the array with `push` and return it.

packages/plugin-api/src/store.ts
```diff
@@ -66,7 +66,7 @@ export interface AllureStore {
 export interface AllureStoreDump {
   testResults: Record<string, TestResult>;
   attachments: Record<string, AttachmentLink>;
-  globalAttachments: AttachmentLink[];
+  globalAttachmentsIds: string[];
```
delatrie: nit:
```suggestion
  globalAttachmentIds: string[];
```

packages/reader-api/src/model.ts
```diff
@@ -2,6 +2,13 @@ export type RawTestStatus = "failed" | "broken" | "passed" | "skipped" | "unknow
 
 export type RawMetadata = { [key: string]: any };
 
+export type RawTestError = {
```
delatrie: nit: perhaps, name it `RawError`? It's not strictly related to a test now.

packages/core/src/store/store.ts
```diff
@@ -194,18 +196,24 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
       this.#globalErrors.push(error);
     });
     this.#realtimeSubscriber?.onGlobalAttachment(async (attachment: ResultFile) => {
-      const attachmentLink: AttachmentLinkFile = {
-        id: md5(attachment.getOriginalFileName()),
+      const ext = attachment.getExtension();
+      // need to generate a unique filename to avoid override of the global files with the same name (such as stdout.txt or stderr.txt)
+      const sourceFileName = `${randomUUID()}.${ext}`;
```
delatrie: update: please, address [this comment](https://github.com/allure-framework/allure3/pull/485#discussion_r2785884739) first.

nit: `ext` could be an empty string, which makes `sourceFileName` end with `.`. Not sure where `originalFileName` is used though, so it might be not a big deal.

packages/reader/src/allure2/index.ts
```diff
@@ -288,6 +302,18 @@ const processTestResultContainer = async (visitor: ResultsVisitor, result: Parti
   }
 };
 
+const processGlobals = async (visitor: ResultsVisitor, globals: Globals) => {
+  const { attachments = [], errors = [] } = globals;
+
+  await visitor.visitGlobals(
+    {
+      attachments: attachments.map((attachment) => convertAttachment(attachment)),
```
delatrie: At this point, `attachments` is not guaranteed to be an array. A part of our strategy for readers is to make them as reliable gatekeepers as possible. We should guard against ill-formed documents producing unexpected or incomplete values in the store.

Here is an example of ensuring an array type:
https://github.com/allure-framework/allure3/blob/42f711278cd404ac4f5c9a6387fc06f6579a136b/packages/reader/src/allure2/index.ts#L207-L209

And the same goes for `errors`.

packages/core/src/report.ts
```diff
@@ -271,7 +271,7 @@ export class AllureReport {
       fixtures,
       attachments: attachmentsLinks,
       environments,
-      globalAttachments = [],
+      globalAttachmentsIds = [],
```
delatrie: nit:
```suggestion
      globalAttachmentIds = [],
```

packages/core/src/store/store.ts
```diff
@@ -194,18 +196,24 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
       this.#globalErrors.push(error);
     });
     this.#realtimeSubscriber?.onGlobalAttachment(async (attachment: ResultFile) => {
-      const attachmentLink: AttachmentLinkFile = {
-        id: md5(attachment.getOriginalFileName()),
+      const ext = attachment.getExtension();
+      // need to generate a unique filename to avoid override of the global files with the same name (such as stdout.txt or stderr.txt)
```
delatrie: Perhaps, we should leave this responsibility to the caller of `sendGlobalAttachment`. All other code that creates `BufferResultFile` instances ensures the names are unique:

  - https://github.com/allure-framework/allure3/blob/42f711278cd404ac4f5c9a6387fc06f6579a136b/packages/reader/src/cucumberjson/index.ts#L234
  - https://github.com/allure-framework/allure3/blob/42f711278cd404ac4f5c9a6387fc06f6579a136b/packages/reader/src/junitxml/index.ts#L184
  - https://github.com/allure-framework/allure3/blob/42f711278cd404ac4f5c9a6387fc06f6579a136b/packages/reader/src/xcresult/xcresulttool/legacy/index.ts#L473

To avoid displaying names like `<UUID>-stdout.txt` in the report, we can add an optional argument to `sendGlobalAttachment` to pass a display name. This will also fully decouple display names from filenames, allowing us to name the attachments, for example, "Standard output" or "Standard output for <id/name of the process>" instead of `stdout.txt`.

packages/reader/src/allure2/index.ts
```diff
@@ -288,6 +302,18 @@ const processTestResultContainer = async (visitor: ResultsVisitor, result: Parti
   }
 };
 
+const processGlobals = async (visitor: ResultsVisitor, globals: Globals) => {
+  const { attachments = [], errors = [] } = globals;
+
+  await visitor.visitGlobals(
+    {
+      attachments: Array.isArray(attachments) ? attachments.map((attachment) => convertAttachment(attachment)) : [],
+      errors,
```
delatrie: `errors` is unchecked here and may contain any value. Since it must be an array of objects, we can use [`isStringAnyRecordArray`](https://github.com/allure-framework/allure3/blob/3fb1961e018a5cf5e6672c7ea6ebcd76fc8911d5/packages/reader/src/xml-utils.ts) to check it:

```js
errors: isStringAnyRecordArray(errors) ? errors : [],
```

PR #492, opened 2026-02-13
Fix fbsu pyramid calculation, rename fbsu age pyramid to status age pyramid

vdvukhzhilov:


packages/web-commons/src/charts/generateStatusAgePyramid.ts
delatrie, line 40: History points are guaranteed to be ordered by their creation time (whether via `.jsonl`, or by Allure Service). Do we really need to sort again? Is it even possible that this sorting gives us a different order? I guess that would only be the case if different history points were generated on different systems with their system clocks out of sync. And in that case, I would still prefer to keep the chronological order, even if the timestamp labels appear confusing.

epszaw (approved):

packages/web-commons/src/charts/generateStatusAgePyramid.ts
```diff
@@ -43,24 +42,43 @@ export const generateFBSUAgePyramid = (props: {
...
-  const [earliestHdp, ...hdps] = limitedHistoryPoints;
+  const hdps = limitedHistoryPoints.map((datapoint) => ({
+    ...datapoint,
+    testResults: Object.values(datapoint.testResults).reduce(
+      (acc, testResult) => {
+        if (!testResult.historyId) {
+          return acc;
+        }
+
+        const isInCurrentRun = testResults.findIndex((tr) => tr.historyId === testResult.historyId) !== -1;
```
delatrie, line 65: This has quadratic complexity (number of test results ^ 2 * number of history points).
Better use a lookup:

```ts
const historyIds = new Set(testResults.map((tr) => tr.historyId));

// ...

const isInCurrentRun = historyIds.has(testResult.historyId);
```

Now it's N*logN.

packages/web-commons/src/charts/generateStatusAgePyramid.ts
delatrie, line 40: A canonical age pyramid is ordered from bottom to top: the bottom line corresponds to the youngest test results, the top one - to the oldest ones.

packages/web-components/src/components/Charts/StatusAgePyramidChartWidget/index.tsx
delatrie, line 78: If all tests are passed, the message tells "No results," which is confusing. Can we customize it to be something like "All tests are passed." in such a case?

delatrie, line 14: How about growing failed and broken bars to the right? Left-to-right flow seems more natural to read and I'd assume people gonna deal with failed/broken tests more oftenly than with skipped/unknow, so it will follow this natural flow in the majority of cases.

Edited: just to clarify, I suggest swapping skipped/unknown with failed/broken

delatrie, line 111: Is it possible to somehow highlight the vertical zero-line? Currently, it's not that apparent if there are few bars on one side (or no bars at all):

Image

vdvukhzhilov: Sure, this was all done in short deadlines, so no time for such things

vdvukhzhilov: Yeah, in vertical graphs it's nonexisting 😢

delatrie: Well, that's a bummer :(

PR #496, opened 2026-02-19
Relative and absolute history modes

epszaw:
enchance allure history methods

### DIFF
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -159,7 +159,7 @@ export class AllureLocalHistory implements AllureHistory {
 
       // in case when limit is undefined – the history is unlimited, so we need to add the point too
       if (limit !== 0) {
-        this.#cachedHistory.unshift(data);
+        this.#cachedHistory.push(data);
       }
 
       if (limit) {
--- a/packages/core/test/history.test.ts
+++ b/packages/core/test/history.test.ts
@@ -501,8 +501,8 @@ describe("AllureLocalHistory", () => {
       const secondRead = await history.readHistory();
 
       expect(secondRead).toEqual([
-        expect.objectContaining({ name: "New entry" }),
         expect.objectContaining({ name: "Entry 1" }),
+        expect.objectContaining({ name: "New entry" }),
       ]);
     });
   });


packages/core/src/history.ts
```diff
@@ -120,19 +130,28 @@ export class AllureLocalHistory implements AllureHistory {
...
+      }
 
-        // Append a new entry; the total number is up to `limit`.
-        const sources = [JSON.stringify(data), Buffer.from([0x0a])];
-        await pipeline(sources, dst);
+      if (limit === 0 && !historyExists) {
+        return;
       }
 
+      if (limit !== 0 && historyExists) {
```
delatrie: nit: `limit !== 0` is always `true` at this point

packages/service/src/service.ts
```diff
@@ -80,12 +80,20 @@ export class AllureServiceClient {
...
    * @param payload
    */
-  async downloadHistory(payload: { branch: string; limit?: number }) {
-    const { branch, limit } = payload;
+  async downloadHistory(payload: { branch?: string; limit?: number }) {
+    const { branch, limit } = payload ?? {};
+    const params = new URLSearchParams();
+
+    if (limit) {
+      params.append("limit", limit.toString());
```
delatrie: `limit` should also be encoded, as it essentially comes from the config and is not guaranteed to be a number at runtime.

packages/core-api/src/history.ts
```diff
@@ -43,6 +43,6 @@ export interface HistoryDataPoint {
  * Provides ability to load and update report history
  */
 export interface AllureHistory {
-  readHistory(branch?: string): Promise<HistoryDataPoint[]>;
-  appendHistory(history: HistoryDataPoint, branch?: string): Promise<void>;
+  readHistory(params?: { branch?: string; limit?: number }): Promise<HistoryDataPoint[]>;
```
delatrie: The `limit` parameter is not used: we only call this method with no arguments from the store, or with the `branch` argument from `createPluginSummary`.

Should we keep it anyway?

packages/core-api/src/history.ts
```diff
@@ -43,6 +43,6 @@ export interface HistoryDataPoint {
...
+  appendHistory(history: HistoryDataPoint, params?: { branch?: string; limit?: number }): Promise<void>;
```
delatrie: The same story as for `readHistory`. The only call we have doesn't pass any arguments:

https://github.com/allure-framework/allure3/blob/8ad1fd80af6385c03eb7a90991e1ab16186d0b8c/packages/core/src/store/store.ts#L249

packages/plugin-api/src/utils/summary.ts
```diff
@@ -1,9 +1,44 @@
...
+  name: string;
+  plugin: string;
+  store: AllureStore;
+  history?: AllureHistory;
+  ci?: CiDescriptor;
+  meta: Record<string, any>;
+}): Promise<PluginSummary> => {
+  const { name, filter, plugin, store, history, meta } = params;
+  const allTrs = await store.allTestResults({ filter });
+  const mainBranchHistory = (await history?.readHistory?.({ branch: "" })) ?? [];
```
delatrie, line 23: In the case of AllureLocalHistory, this call loads and parses the entire history again for each plugin instance that passes history to createPluginSummary.
For AllureServiceHistory, the same applies if we're currently on the default branch. This is especially painful if we have multiple Awesome reports (e.g., with different filters). My estimates show up to +30-50% to the total generation time for 5 reports, 2000 tests each, with 30 history points due to these extra loadings.

History loading is a notable contributor to the total generation time. We should avoid unnecessary calls where possible. For example, we can cache entries in AllureLocalHistory (one-time) or AllureServiceHistory (on a per-branch basis).

packages/plugin-api/src/utils/summary.ts
```diff
@@ -1,9 +1,44 @@
...
+  const newTrs = await store.allNewTestResults(filter, mainBranchHistory);
```
delatrie: `mainBranchHistory` is `[]` for each plugin that calls `createPluginSummary` without `history` (like `dashboard`). That will make all test results new, and the summary will show misleading numbers.

delatrie, line 24: Currently, `allNewTestResults` returns an empty array when called with an empty history. The summary will show "0 new tests," even though all the tests are new.

packages/plugin-dashboard/src/plugin.ts
```diff
@@ -56,29 +55,15 @@ export class DashboardPlugin implements Plugin {
...
       plugin: "Dashboard",
-      newTests: newTrs.map(convertToSummaryTestResult),
-      flakyTests: flakyTrs.map(convertToSummaryTestResult),
-      retryTests: retryTrs.map(convertToSummaryTestResult),
       meta: {
         reportId: context.reportUuid,
         singleFile: this.options.singleFile ?? false,
       },
-    };
+      filter: this.options.filter,
```
delatrie, line 65: Some plugins, like `dashboard`, don't pass `history` to `createPluginSummary`. Is this intentional? This will make the GitHub action always report 0 new tests for their rows in the Allure Report Summary of the PR.

packages/core-api/src/history.ts
epszaw: You're right. I'd removed related logic, but didn't remove the limit.

packages/plugin-api/src/utils/summary.ts
epszaw: Changed the logic and now, the `allNewTestResults` returns empty array only when history doesn't exist, so we can't treat tests as new.

If history exists or empty history is given – all tests are new.

packages/plugin-dashboard/src/plugin.ts
epszaw: Added the parameter to the plugins. Even they don't utilize transitioning statuses functionality, it isn't something bad :)

packages/core/src/history.ts
```diff
@@ -120,24 +129,42 @@ export class AllureLocalHistory implements AllureHistory {
...
+
       if (historyExists) {
         await historyFile.truncate(dst.bytesWritten);
       }
     } finally {
       await historyFile.close();
+
+      // in case when limit is undefined – the history is unlimited, so we need to add the point too
+      if (limit !== 0) {
+        this.#cachedHistory.unshift(data);
```
delatrie: This array is ordered from earliest to most recent entries. The appended entry should go to the end, not to the beginning.

Anyway, it doesn't affect much, because there is almost nothing left to happen after appendHistory is called

epszaw: 👍

PR #500, opened 2026-02-20
fix posix paths

### LINKED ISSUE #477: singleFile: true with allure-cypress does not generate widgets/default/tree.json
When using `singleFile: true` with Allure 3 + `allure-cypress` + Awesome plugin, the generated report is missing the `widgets` structure, and the HTML fails to load data with the error:

> Data "widgets/default/tree.json" not found!

With `singleFile: false` the same setup generates a fully working report.

---

### Environment

- Allure CLI: 3.1.0  
- allure-cypress: 3.4.5  
- Cypress: 15.6.0  
- Node.js: <your version here>  
- OS: <Windows 10 / Windows 11 / other>  

Project structure (Cypress root):

- `results/reports/allure-results` → contains multiple `*-result.json`, `environment.properties`, `history.jsonl`, etc.  
- `results/reports/allure-report` → Allure output folder  

Config file at: `src/cypress/allurerc.mjs`

```ts

export default defineConfig({
  name: "Relatório de Execução de Testes E2E - Eyon Telecom",
  output: "./results/reports/allure-report",
  historyPath: "./results/reports/history.jsonl",
  appendHistory: true,
  plugins: {
    awesome: {
      options: {
        singleFile: true,
        reportLanguage: "pt-br",
      },
    },
  },
});
```
---

### Steps to Reproduce

1. Run Cypress tests with `allure-cypress` enabled, which generates results under:

   ```text
   results/reports/allure-results
   ```

2. From the Cypress project directory (where `package.json` and `allurerc.mjs` live), generate the report:

   ```bash
   allure generate results/reports/allure-results --config ./allurerc.mjs
   ```
3. Open the report:

   ```bash
   allure open results/reports/allure-report
   ```

4. The browser opens `index.html` under `results/reports/allure-report`.

---

### Actual Result

- In `results/reports/allure-report` only these files are generated:

  ```text
  index.html
  summary.json
  ```

- There is **no** `widgets` directory, so paths like `widgets/default/tree.json` do not exist.
- When opening the report in a browser, the UI shows:

  ```text
  Data "widgets/default/tree.json" not found!
  ```

Image

- If I set `singleFile: false` in `allurerc.mjs` and regenerate, the report is created correctly with the `widgets` folder and loads fine.

---

### Expected Result

- With `singleFile: true`, I expect Allure to generate a working single-file report (or at least the same widget data structure as the non-single-file mode), so that opening `index.html` does **not** result in `Data "widgets/default/tree.json" not found!`.
- Ideally, the Awesome plugin should:
  - Produce a valid standalone report, or
  - Fail with a clear error message if `singleFile` cannot be applied, instead of generating a partial report without `widgets`.

---

### Notes

- The exact same project, config, and Allure versions work correctly when `singleFile` is disabled.
- This behavior looks specific to the Awesome plugin with `singleFile: true` in combination with `allure-cypress`.

formaceft-93:
Fix close #477

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -15,6 +15,9 @@ export {
   maxFailuresRule,
   minTestsCountRule,
   successRateRule,
+  maxDurationRule,
+  allTestsContainEnvRule,
+  environmentsTestedRule,
   convertQualityGateResultsToTestErrors,
   stringifyQualityGateResults,
 } from "./qualityGate/index.js";
--- a/packages/core/src/qualityGate/qualityGate.ts
+++ b/packages/core/src/qualityGate/qualityGate.ts
@@ -99,6 +99,7 @@ export class QualityGate {
           },
           expected,
           knownIssues,
+          environment: environment || DEFAULT_ENVIRONMENT,
         });
 
         if (result.success) {
--- a/packages/core/src/qualityGate/rules.ts
+++ b/packages/core/src/qualityGate/rules.ts
@@ -68,4 +68,55 @@ export const maxDurationRule: QualityGateRule<number> = {
   },
 };
 
-export const qualityGateDefaultRules = [maxFailuresRule, minTestsCountRule, successRateRule, maxDurationRule];
+/**
+ * Fails if any test in the run does not have the given environment.
+ * Expected: environment name (string).
+ */
+export const allTestsContainEnvRule: QualityGateRule<string> = {
+  rule: "allTestsContainEnv",
+  message: ({ actual, expected }) =>
+    `Not all tests contain the required environment, ${bold(`"${expected}"`)}; ${bold(actual.length === 1 ? "one" : String(actual))} ${actual.length === 1 ? "test has" : "tests have"} different or missing environment`,
+  validate: async ({ trs, expected }) => {
+    const testsWithoutEnv = trs.filter((tr) => (tr.environment ?? "") !== expected);
+    const actual = testsWithoutEnv.length;
+
+    return {
+      success: actual === 0,
+      actual,
+    };
+  },
+};
+
+/**
+ * Fails if the run does not contain at least one test for each of the given environments.
+ * Expected: array of environment names (string[]).
+ */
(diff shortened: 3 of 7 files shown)


packages/plugin-classic/src/writer.ts
```diff
@@ -104,7 +111,7 @@ export class ReportFileDataWriter implements ClassicDataWriter {
       return;
     }
 
-    await this.reportFiles.addFile(joinPosix("data", "attachments", source), contentBuffer);
+    await this.reportFiles.addFile(joinPosix("data", "attachments", getPosixPath(source)), contentBuffer);
```
epszaw: I think, we can go a bit further and create our own `join` and `resolve` path utilities with posix conversion under the hood. It seems, in that way, we'd have fewer opportunities to make mistakes.

For example:

```js
joinPosix("data", "attachments", source) // already does the conversion of `source`
```

packages/web-commons/src/data.ts
```diff
@@ -20,11 +22,22 @@ export const loadReportData = async (name: string): Promise<string> => {
...
-    if (globalThis.allureReportData[name]) {
-      return resolve(globalThis.allureReportData[name] as string);
-    } else {
-      return reject(new Error(`Data "${name}" not found!`));
+    const dataByName = globalThis.allureReportData ?? {};
+    if (dataByName[name]) {
+      return resolve(dataByName[name] as string);
+    }
+
+    const windowsKey = getPosixPath(name).replace(/\//g, "\\");
```
epszaw: We have to use posix path as keys on any OS, even on windows because the report can be generated on any OS and opened on any OS. So better to prevent situations when the report data object contain reverse-slash keys.

packages/core-api/src/utils/path.ts
```diff
@@ -0,0 +1,16 @@
+export const toPosixPath = (path: string) => path.replace(/\\/g, "/");
+
+export const joinReportPath = (...parts: string[]) => {
```
epszaw: Let's rename to `joinPosixPath` because there's no any relations with the report itself.

Issue #424, opened 2025-12-23
Improve stability visuals


delatrie:
#187 defines a metric called "Stability score:" 

## <a name="stability-score"></a>Stability score definition

A stability score is defined at the test level. The idea is to assign a number between 0 and 1 to a test based on the number of status transitions it has.

Let `Nss > 0` be the history depth we're considering for stability score calculations.
Let `0 < Nst <= Nss` be a stabilization period that resets the stability score.

Then, the following procedure should be applied to get the stability score of the test:

1. Append the current status with the historic ones to form a sequence of statuses.
2. Filter out statuses other than passed, failed, and broken.
3. Take the most recent `Nss` statuses. If there are fewer than `Nss` statuses, take all of them.
   Let `Nsse` be `min(N, Nss)`.
   **Rule 1**: If `Nsse` is zero, the stability score is **undefined**.
   **Rule 2**: If `Nsse` is one, the stability score is **1**.
4. Apply the stabilization rule:
   Find the latest `Nst` consecutive statuses.
   **Rule 3**: If these statuses are the most recent ones in the whole sequence, the stability score is **1**.
   Otherwise, let `Ist` be the number of the first edge that follows the most recent such status, `1 <= Ist <= Nsse - 1`.
   If no such statuses exist, `Ist` is `1`.
5. Apply the tolerance rules:
     - **Rule 4**: If the edges `[Ist, Nsse - 1]` contain only one transition, the stability score is **1**.
     - **Rule 5**: If the edges `[Ist, Nsse - 1]` contain only two transitions (not necessarily consecutive): P->B followed by B->P, the stability score is **1**.
     - **Rule 6**: If the edges `[Ist, Nsse - 1]` contain only two transitions (not necessarily consecutive): P->F followed by F->P, the stability score is **1**.
6. Calculate the stability rate using the formula:
   **Rule 7**: `SR = SUM(SRi)`, where `i` ranges from `Ist` to `Nsse - 1` (both ends included), and `SRi` is defined as follows: 
     - If the edge number `i` is a transition, then:
         `SRi = i / (1 + 2 + ... + Nsse-1)`
     - Otherwise, `SRi = 0`

The motivation for the stabilization rule is to get fixed tests (which are no longer flaky) off the radar sooner.
The motivation for the tolerance rules is to avoid marking stable tests as unstable. For example:

  - regression detected by a test, then fixed: `P -> F -> P`
  - a new failed test was introduced, then the feature got implemented: `F -> P`

[...]

PR #511, opened 2026-02-25
chore: fix charts

vdvukhzhilov:
Closes https://github.com/allure-framework/allure3/issues/424

### DIFF
--- a/packages/web-commons/src/charts/generateStabilityDistributionChart.ts
+++ b/packages/web-commons/src/charts/generateStabilityDistributionChart.ts
@@ -235,6 +235,10 @@ export const generateStabilityDistributionChart = (props: {
       continue;
     }
 
+    if (!isSignificantStatus(tr.status)) {
+      continue;
+    }
+
     if (skipStatuses.has(tr.status)) {
       continue;
     }


packages/web-commons/src/charts/generateStabilityDistributionChart.ts
```diff
@@ -4,33 +4,189 @@ import type {
...
+    return undefined;
+  }
+  // Rule 2: single status → fully stable
+  if (effectiveSequenceLength === 1) {
+    return 1;
+  }
+
+  const stabilBlock = findLatestConsecutiveSameBlock(statuses, effectiveSequenceLength, stabilizationPeriod);
+
+  // Rule 3: that block is at the very end and all "passed" → test is stable
```
delatrie: Rule 3 is not specific to `passed`. It also applies to `failed` and `broken`. The motivation is that a false-positive flaky test may get fixed, and reliably show the problem from now on. In such a case, we want to quickly capture the fact that it's stable now.

packages/web-commons/src/charts/generateStabilityDistributionChart.ts
```diff
@@ -4,33 +4,189 @@ import type {
...
+    if (statuses[i - 1] === statuses[i]) {
+      continue;
+    }
+    transitionCount++;
+    if (transitionCount === 1) {
+      firstTransitionIdx = i;
+    } else if (transitionCount === 2) {
+      secondTransitionIdx = i;
+    }
+    if (!(statuses[i - 1] === "failed" && statuses[i] === "passed")) {
```
delatrie: There was an error in my example 6 in #424 (I had changed the input at some point, but didn't update the values properly). `SR6` must be `6/45`, not `0`, and the resulting score, therefore, must be `36/45 = 0.8`.

The procedure defines nothing special about the `failed -> passed` transition at step 6.

Sorry for confusing you.

packages/web-commons/src/charts/generateStabilityDistributionChart.ts
```diff
@@ -4,33 +4,189 @@ import type {
...
+  // First edge (1-based) to consider: right after the stabil block, or 1 if no block
+  const firstEdgeAfterStabilBlock = stabilBlock !== null ? stabilBlock.endIndex + 1 : 1;
+  const lastEdge = effectiveSequenceLength - 1;
+  const denominator = ((effectiveSequenceLength - 1) * effectiveSequenceLength) / 2;
+
+  const { transitionCount, firstTransitionIdx, secondTransitionIdx, instabilitySum } = computeTransitionsInRange(
+    statuses,
+    firstEdgeAfterStabilBlock,
+    lastEdge,
+    denominator,
```
delatrie: nit: a slightly more simple alternative is to sum the numerators only and apply the denominator at the and:

```js
return instabilitySum / denominator;
```

packages/web-commons/src/charts/generateStabilityDistributionChart.ts
```diff
@@ -4,34 +4,187 @@ import type {
...
+ */
+const getStatusSequence = (historyDataPoints: HistoryDataPoint[], tr: TestResult): TestStatus[] => {
+  let block: TestStatus[] = [];
+
+  for (const hdp of historyDataPoints) {
+    const htr = hdp.testResults[tr.historyId!];
+
+    if (!htr) {
+      // Gap: test was not in this run — keep only statuses after this point
+      block = [];
```
delatrie, line 173: I doubt we need such a reset logic. Often, missing tests are essentially the same as skipped. For example, skipped NUnit tests are missing by default unless `[AllureDisplayIgnored]` is applied. Let's just treat gaps as skipped test results.

PR #512, opened 2026-02-25
feat: add new quality gate rules

vdvukhzhilov:
Closes https://github.com/allure-framework/allure-org/issues/305
Closes https://github.com/allure-framework/allure-org/issues/304

### DIFF
--- a/packages/core/src/qualityGate/rules.ts
+++ b/packages/core/src/qualityGate/rules.ts
@@ -95,10 +95,14 @@ export const environmentsTestedRule: QualityGateRule<string[]> = {
   rule: "environmentsTested",
   message: ({ actual, expected }) =>
     `The following environments were not tested: "${actual.join('", "')}"; expected all of: "${expected.join('", "')}"`,
-  validate: async ({ trs, expected }) => {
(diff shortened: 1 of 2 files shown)


packages/core/src/qualityGate/rules.ts
```diff
@@ -68,4 +68,51 @@ export const maxDurationRule: QualityGateRule<number> = {
...
+
+/**
+ * Fails if the run does not contain at least one test for each of the given environments.
+ * Expected: array of environment names (string[]).
+ */
+export const environmentsTestedRule: QualityGateRule<string[]> = {
+  rule: "environmentsTested",
+  message: ({ actual, expected }) =>
+    `The following environments were not tested: "${actual.join('", "')}"; expected all of: "${expected.join('", "')}"`,
+  validate: async ({ trs, expected }) => {
```
epszaw: Use state to accumulate already tested environments.

Quality gate can also work in real-time mode, when it's receiving multiple batch of test results, so we can't fail the run in this case util we've processed all the results.

`allTestsContainEnvRule` will work without state because we need to ensure that the given environment is present in every single test result, though.

PR #522, opened 2026-03-02
add env validation

formaceft-93:
Add validation for environment 

Close [#31](https://github.com/allure-framework/allure-org/issues/310)

### DIFF
--- a/packages/web-awesome/test/components/EnvironmentPicker.test.tsx
+++ b/packages/web-awesome/test/components/EnvironmentPicker.test.tsx
@@ -106,13 +106,12 @@ describe("components > EnvironmentPicker", () => {
     expect(queryTooltip()).not.toBeInTheDocument();
 
     setNarrowViewport(true);
-    fireEvent(window, new Event("resize"));
-
-    const tooltipTrigger = getTooltipTrigger();
     await waitFor(() => {
-      expect(tooltipTrigger.childElementCount).toBe(2);
+      fireEvent(window, new Event("resize"));
+      expect(getTooltipTrigger().childElementCount).toBe(2);
     });
 
+    const tooltipTrigger = getTooltipTrigger();
     fireEvent.mouseEnter(tooltipTrigger);
     await waitFor(() => {
       expect(queryTooltip()).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


packages/core/src/store/store.ts
```diff
@@ -90,6 +100,46 @@ const hidePreviousAttempt = (state: Map<string, Map<string, TestResult>>, testRe
...
+  const validationResult = validateEnvironmentName(name);
+
+  if (!validationResult.valid) {
+    return DEFAULT_ENVIRONMENT;
+  }
+
+  return validationResult.normalized;
+};
+
+const normalizeEnvironmentQueryOrThrow = (env: string, source: string): string => {
```
epszaw: The function does pretty the same job as `coerceRestoredEnvironmentName`, so we can go further with a single one

packages/core/src/store/store.ts
```diff
@@ -90,6 +100,46 @@ const hidePreviousAttempt = (state: Map<string, Map<string, TestResult>>, testRe
...
+  environmentsConfig: EnvironmentsConfig,
+): string => {
+  if (typeof testResult.environment === "string") {
+    return coerceRestoredEnvironmentName(testResult.environment);
+  }
+
+  return forcedEnvironment ?? matchEnvironment(environmentsConfig, testResult);
+};
+
+const environmentKeyForQualityGateResultIndexing = (environment: string | undefined): string => {
```
epszaw: The function also does the same job as `coerceRestoredEnvironmentName`, so let's simplify the code too

packages/web-awesome/src/components/MetadataButton/index.tsx
```diff
@@ -1,33 +1,79 @@
...
   setIsOpen,
   counter,
   title,
+  titleTooltipText,
+  truncateTitle = false,
   ...rest
 }) => {
-  return (
+  const titleRef = useRef<HTMLSpanElement>(null);
+  const [isTitleTruncated, setIsTitleTruncated] = useState(false);
```
epszaw: This hook with the resize observer is pretty the same as [here](https://github.com/allure-framework/allure3/pull/522/changes#diff-08bb9a66d9225e1ac72057b8fe31858d4891b1e8d00d490cfafca115bd708f9fR44). I think, we can move the logic to a hook and re-use across the components.

What do you think about it?

PR #532, opened 2026-03-05
Add hidden label feature

formaceft-93:
Added hideLabels support (string | RegExp):
- by default, all labels starting with _ are hidden.

Closes https://github.com/allure-framework/allure-org/issues/316 , https://github.com/allure-framework/allure-org/issues/315

### DIFF
--- a/packages/service/src/history.ts
+++ b/packages/service/src/history.ts
@@ -1,41 +1,7 @@
-import type { AllureHistory, HistoryDataPoint } from "@allurereport/core-api";
+import { type AllureHistory, normalizeHistoryDataPointUrls } from "@allurereport/core-api";
 
 import type { AllureServiceClient } from "./service.js";
 import { KnownError } from "./utils/http.js";
-
-const normalizeHistoryDataPointUrls = (historyDataPoint: HistoryDataPoint): HistoryDataPoint => {
-  const { url } = historyDataPoint;
-
-  if (!url) {
-    return historyDataPoint;
-  }
-
-  let testResults = historyDataPoint.testResults;
-
-  for (const [historyId, historyTestResult] of Object.entries(historyDataPoint.testResults)) {
-    if (historyTestResult.url) {
-      continue;
-    }
-
-    if (testResults === historyDataPoint.testResults) {
-      testResults = { ...historyDataPoint.testResults };
-    }
-
-    testResults[historyId] = {
-      ...historyTestResult,
-      url,
-    };
-  }
-
-  if (testResults === historyDataPoint.testResults) {
-    return historyDataPoint;
-  }
-
-  return {
-    ...historyDataPoint,
-    testResults,
-  };
-};
 
 export class AllureRemoteHistory implements AllureHistory {
   constructor(readonly params: { allureServiceClient: AllureServiceClient; limit?: number; branch?: string }) {}
--- a/packages/core-api/src/utils/history.ts
+++ b/packages/core-api/src/utils/history.ts
@@ -61,6 +61,40 @@ export const filterUnknownByKnownIssues = (
   });
 };
 
+export const normalizeHistoryDataPointUrls = (historyDataPoint: HistoryDataPoint): HistoryDataPoint => {
+  const { url } = historyDataPoint;
+
+  if (!url) {
+    return historyDataPoint;
+  }
+
+  let testResults = historyDataPoint.testResults;
+
+  for (const [historyId, historyTestResult] of Object.entries(historyDataPoint.testResults)) {
+    if (historyTestResult.url) {
+      continue;
+    }
+
+    if (testResults === historyDataPoint.testResults) {
+      testResults = { ...historyDataPoint.testResults };
+    }
+
+    testResults[historyId] = {
+      ...historyTestResult,
+      url,
+    };
+  }
+
+  if (testResults === historyDataPoint.testResults) {
+    return historyDataPoint;
+  }
+
+  return {
+    ...historyDataPoint,
+    testResults,
+  };
+};
+
 export const selectHistoryTestResults = (
   historyDataPoints: HistoryDataPoint[],
   historyIdCandidates: readonly string[],
--- a/packages/core-api/test/utils/history.test.ts
+++ b/packages/core-api/test/utils/history.test.ts
@@ -7,6 +7,7 @@ import {
   filterUnknownByKnownIssues,
   getFallbackHistoryId,
   getHistoryIdCandidates,
+  normalizeHistoryDataPointUrls,
   selectHistoryTestResults,
   stringifyHistoryParams,
 } from "../../src/utils/history.js";
@@ -159,5 +160,30 @@ describe("history utils", () => {
 
     expect(selectedHistoryTestResult).toBe(historyTestResult);
     expect(selectedHistoryTestResult.url).toBe("https://history");
+  });
+
+  it("should normalize nested history urls from datapoint url when needed", () => {
+    const historyTestResult = { id: "primary", name: "primary", status: "passed", url: "" };
+    const historyDataPoint = {
+      uuid: "first",
+      name: "Entry 1",
+      timestamp: 1,
+      knownTestCaseIds: [],
+      metrics: {},
+      url: "https://history",
+      testResults: {
+        primary: historyTestResult,
+      },
+    };
+
+    expect(normalizeHistoryDataPointUrls(historyDataPoint)).toEqual({
+      ...historyDataPoint,
+      testResults: {
+        primary: {
+          ...historyTestResult,
+          url: "https://history",
+        },
+      },
+    });
   });
 });
--- a/packages/core/src/history.ts
+++ b/packages/core/src/history.ts
@@ -4,43 +4,16 @@ import path from "node:path";
 import readline from "node:readline/promises";
 import { pipeline } from "node:stream/promises";
 
-import type { AllureHistory, HistoryDataPoint, HistoryTestResult, TestCase, TestResult } from "@allurereport/core-api";
+import {
+  type AllureHistory,
+  type HistoryDataPoint,
+  type HistoryTestResult,
+  normalizeHistoryDataPointUrls,
+  type TestCase,
+  type TestResult,
+} from "@allurereport/core-api";
 
 import { isFileNotFoundError } from "./utils/misc.js";
-
-const normalizeHistoryDataPointUrls = (historyDataPoint: HistoryDataPoint): HistoryDataPoint => {
-  const { url } = historyDataPoint;
-
-  if (!url) {
-    return historyDataPoint;
-  }
-
-  let testResults = historyDataPoint.testResults;
-
-  for (const [historyId, historyTestResult] of Object.entries(historyDataPoint.testResults)) {
-    if (historyTestResult.url) {
-      continue;
-    }
-
-    if (testResults === historyDataPoint.testResults) {
-      testResults = { ...historyDataPoint.testResults };
-    }
-
-    testResults[historyId] = {
-      ...historyTestResult,
-      url,
-    };
-  }
-
-  if (testResults === historyDataPoint.testResults) {
-    return historyDataPoint;
-  }
-
-  return {
-    ...historyDataPoint,
-    testResults,
-  };
-};
 
 const createHistoryItems = (testResults: TestResult[], remoteUrl: string) => {
   return testResults


packages/plugin-awesome/README.md
```diff
@@ -56,4 +56,5 @@ The plugin accepts the following options:
 | `reportLanguage` | Default language of the report                                                                                                                              | `string`                                                     | OS language                   |
 | `ci`           | CI data which will be rendered in the report                                                                                                                | `{ type: "github" \| "jenkins", url: string, name: string }` | `undefined`                   |
 | `groupBy`      | Grouping tests by labels or combining labels. By default, tests are grouped using the `titlePath` provided by the test framework. | `string`                                                      | `[]`(Grouping by `titlepath`) |
+| `hideLabels`   | Hides labels by name in Awesome report data and UI. Accepts exact names and regular expressions. Labels with names that start with `_` are hidden by default. | `(string \| RegExp)[]`                                        | `[]` + default `_` prefix rule |
```
epszaw: Add `--hide-labels` option to all commands which produce any kind of reports to be able to use the command in this way:

```shell
allure run --hide-labels=foo --hide-labels=bar -- npm t
allure generate --hide-labels=foo --hide-labels=bar ./allure-results
allure awesome --hide-labels=foo --hide-labels=bar ./allure-results
```

packages/plugin-awesome/src/converters.ts
```diff
@@ -11,6 +11,28 @@ import MarkdownIt from "markdown-it";
...
+
+    if (matcher.global || matcher.sticky) {
+      matcher.lastIndex = 0;
+    }
+
+    return matcher.test(labelName);
+  });
+};
+
+const filterLabels = (labels: TestLabel[], hideLabels: readonly (string | RegExp)[] = []): TestLabel[] => {
```
epszaw: Move these functions to the `core` because it'll be use across many packages

packages/core/src/qualityGate/rules.ts
```diff
@@ -1,14 +1,26 @@
-import { filterSuccessful, filterUnsuccessful } from "@allurereport/core-api";
-import { type QualityGateRule } from "@allurereport/plugin-api";
+import { type TestResult, filterSuccessful, filterUnsuccessful } from "@allurereport/core-api";
+import { type QualityGateRule, getHistoryIdCandidates } from "@allurereport/plugin-api";
 import { bold } from "yoctocolors";
 
+const filterUnknownByKnownIssues = (trs: TestResult[], knownIssueHistoryIds: ReadonlySet<string>): TestResult[] => {
```
epszaw: Expose the function by exporting it

packages/web-commons/test/charts/generators.test.ts
```diff
@@ -0,0 +1,173 @@
...
+  return {
+    allEnvironments: async () => [DEFAULT_ENVIRONMENT],
+    allTestResults: async () => testResults,
+    allHistoryDataPoints: async () => historyDataPoints,
+    testsStatistic: async () => statistic ?? { total: testResults.length, failed: testResults.length },
+  } as unknown as AllureStore;
+};
+
+describe("generateCharts", () => {
+  it("maps fallback history alias to current historyId for status age pyramid", async () => {
```
epszaw: We're following naming convention for the tests: `it("should map ...`

packages/plugin-awesome/test/converters.test.ts
```diff
@@ -57,8 +57,111 @@ describe("convertTestResult", () => {
...
+        ],
+      }),
+    );
+
+    expect(result.labels).toEqual([{ name: "tag", value: "api" }]);
+    expect(result.groupedLabels._internal).toBeUndefined();
+    expect(result.groupedLabels.tag).toEqual(["api"]);
+  });
+
+  it("keeps non-hidden labels when hideLabels is configured", () => {
```
epszaw: The test does exact the same check as [this one](https://github.com/allure-framework/allure3/pull/532/changes#diff-649ef255a297e916735526461d457f2f67008314210a1f6a1195a045962d74f4R65)

packages/plugin-api/src/utils/historyIds.ts
```diff
@@ -0,0 +1,52 @@
+import {
+  type TestLabel,
+  type TestParameter,
+  fallbackTestCaseIdLabelName,
+  findLastByLabelName,
+} from "@allurereport/core-api";
+
+import { md5 } from "./misc.js";
+
+type TestResultLike = {
```
epszaw: IMO better to pick properties from the original type `Pick<TestResult, "labels" | "parameters" | "historyId">`

packages/core/src/store/store.ts
```diff
@@ -623,7 +625,38 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
+        }
+
+        if (dp.url) {
+          const url = new URL(dp.url);
+
+          url.hash = tr.id;
+
+          acc.push({
+            ...htr,
+            url: url.toString(),
```
epszaw: I'm not sure, that we need to perform any history data point transformations there, but just filter them

packages/web-commons/src/charts/generators.ts
```diff
@@ -21,6 +21,81 @@ import { generateTrSeveritiesChart } from "./generateTrSeveritiesChart.js";
...
+    for (const [legacyHistoryId, primaryHistoryId] of aliases) {
+      if (!Object.prototype.hasOwnProperty.call(testResults, legacyHistoryId)) {
+        continue;
+      }
+
+      if (Object.prototype.hasOwnProperty.call(testResults, primaryHistoryId)) {
+        continue;
+      }
+
+      if (!changed) {
```
epszaw: Let's remove the flag and copy the results on line 64. Now it looks like an unnecessary step

packages/core/src/config.ts
```diff
@@ -285,6 +286,31 @@ export const getPluginInstance = (config: FullConfig, predicate: (plugin: Plugin
   return config?.plugins?.find(predicate);
 };
 
+export const applyHideLabelsToAwesomePlugins = (config: FullConfig, hideLabels?: readonly string[]): FullConfig => {
```
epszaw: The `core` package includes abstract things which can be used across all the packages inside `allure3`. The current implementation binds the function strictly to awesome plugin, but it shouldn't. Move the function to awesome plugin or make it more abstract and generic.

packages/cli/src/commands/generate.ts
```diff
@@ -75,6 +79,7 @@ export class GenerateCommand extends Command {
       port: this.port,
       historyLimit: this.historyLimit ? parseInt(this.historyLimit, 10) : undefined,
     });
+    applyHideLabelsToAwesomePlugins(config, this.hideLabels);
```
epszaw: Once you use the function in `cli` package only, I propose to move the function to the `cli` package. In that way, the function would exist in a top-level package which does know about all the available plugins and can manipulate the config file.

epszaw: I think, we can move `hideLabels` out of awesome plugin to the configuration because it can be used in future by any other plugin.

If you move it, you would be able to use override argument of `readConfig` function without additional logic around it.

README.md
```diff
@@ -144,6 +151,7 @@ The configuration file allows you to fine-tune report generation. Key options in
 
 - **`name`**: Specifies the report’s display name.
 - **`output`**: Defines the directory where the report will be saved.
+- **`hideLabels`** *(`(string | RegExp)[]`)*: Hides matching labels by name in report data. Awesome respects this option today. Labels with names starting with `_` are hidden by default.
```
epszaw:
```suggestion
- **`hideLabels`** *(`(string | RegExp)[]`)*: Hides matching labels by name in report data. Currently, only Allure Awesome report respects the option. Labels with names starting with `_` are hidden by default.
```

README.md
```diff
@@ -50,6 +50,12 @@ For example, if you're using `npm` as your test runner, the command would be:
 npx allure run -- npm test
 ```
 
+If you need to hide some labels in the generated Awesome report, repeat `--hide-labels`:
```
epszaw:
```suggestion
If you need to hide some labels use `--hide-labels` option:
```

packages/core-api/src/utils/label.ts
```diff
@@ -12,3 +12,26 @@ export const findLastByLabelName = (labels: TestLabel[], name: string): string |
...
+  if (labelName.startsWith("_")) {
+    return true;
+  }
+
+  return matchers.some((matcher) => {
+    if (typeof matcher === "string") {
+      return matcher === labelName;
+    }
+
+    matcher.lastIndex = 0;
```
epszaw: It looks like a thing that does nothing

packages/core-api/src/utils/label.ts
```diff
@@ -12,3 +12,26 @@ export const findLastByLabelName = (labels: TestLabel[], name: string): string |
...
+
+    return matcher.test(labelName);
+  });
+};
+
+export const filterLabels = <T extends Pick<TestLabel, "name">>(
```
epszaw: Once the function has a single predicate, let's use `shouldHideLabel` directly. Currently, the function is a just straightforward wrapper without any additional logic around the filtering. Let's remove the function in favour of using just `shouldHideLabel` one

packages/cli/README.md
```diff
@@ -140,6 +147,7 @@ The configuration file allows you to fine-tune report generation. Key options in
 
 - **`name`**: Specifies the report’s display name.
 - **`output`**: Defines the directory where the report will be saved.
+- **`hideLabels`** *(`(string | RegExp)[]`)*: Hides matching labels by name in report data. Awesome respects this option today. Labels with names starting with `_` are hidden by default.
```
epszaw:
```suggestion
- **`hideLabels`** *(`(string | RegExp)[]`)*: Hides matching labels by name in report data. Currently, only Allure Awesome report respects the option. Labels with names starting with `_` are hidden by default.
```

packages/cli/README.md
```diff
@@ -50,6 +50,12 @@ For example, if you're using `npm` as your test runner, the command would be:
 npx allure run -- npm test
 ```
 
+To hide specific labels in the generated Awesome report, repeat `--hide-labels`:
```
epszaw:
```suggestion
To hide specific labels use `--hide-labels` option:
```

packages/cli/test/commands/awesome.test.ts
```diff
@@ -171,6 +172,37 @@ describe("awesome command", () => {
...
+      hideLabels: ["owner", "tag"],
+    });
+    const reportConfig = (AllureReport as Mock).mock.calls[0]?.[0];
+
+    expect(reportConfig.plugins).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          id: "awesome",
+          options: expect.not.objectContaining({
+            hideLabels: expect.anything(),
```
epszaw: Check exact value instead

packages/cli/test/commands/run.test.ts
```diff
@@ -1,8 +1,116 @@
...
 
 import { RunCommand } from "../../src/commands/run.js";
 
+const { exitMock, processStream, AllureReportMock } = vi.hoisted(() => {
+  const exitMock = vi.fn();
+  const processStream = {
+    setEncoding: vi.fn().mockReturnThis(),
+    on: vi.fn().mockReturnThis(),
+  };
+  const AllureReportMock = vi.fn(function (this: Record<string, unknown>, config: unknown) {
```
epszaw: We already have `AllureReportMock` implementation in `packages/cli/test/utils.ts`, use it instead.

packages/core/src/qualityGate/rules.ts
```diff
@@ -1,14 +1,29 @@
-import { filterSuccessful, filterUnsuccessful } from "@allurereport/core-api";
-import { type QualityGateRule } from "@allurereport/plugin-api";
+import { type TestResult, filterSuccessful, filterUnsuccessful } from "@allurereport/core-api";
+import { type QualityGateRule, getHistoryIdCandidates } from "@allurereport/plugin-api";
 import { bold } from "yoctocolors";
 
+export const filterUnknownByKnownIssues = (
```
epszaw: The function can be useful across the project, so let's move it to `packages/core-api/src/utils/history.ts` and use it out of there

packages/plugin-api/src/utils/historyIds.ts
```diff
@@ -0,0 +1,48 @@
+import {
```
epszaw: What do you think to move these utils to `packages/core-api/src/utils/history.ts` too?

epszaw (changes requested):

packages/service/src/history.ts
```diff
@@ -1,8 +1,42 @@
-import type { AllureHistory } from "@allurereport/core-api";
+import type { AllureHistory, HistoryDataPoint } from "@allurereport/core-api";
 
 import type { AllureServiceClient } from "./service.js";
 import { KnownError } from "./utils/http.js";
 
+const normalizeHistoryDataPointUrls = (historyDataPoint: HistoryDataPoint): HistoryDataPoint => {
```
epszaw: The utility looks pretty the same as in `packages/core/src/history.ts`. Reuse the core one here

epszaw (approved):

PR #536, opened 2026-03-10
Add id to envs

formaceft-93:
### Summary

- Added stable environment identity support.
- environmentId is now the canonical machine key, environment is only the display name.
- Internal environment logic is now ID-first: history, retries, quality gate, report grouping, UI selection, and TestOps uploads.

Explicit IDs are now supported and are the recommended way to configure environments:
```
environments: {
  staging_eu: {
    name: "Staging EU",
    matcher: ...
  }
}
```

### Legacy behavior

- For legacy and synthetic environments, the system now generates a machine-safe ID from the normalized environment name.
- This improves environmentId consistency, but does not preserve continuity after rename.

### Migration

- To keep continuity after rename, the environment should be configured with an explicit stable ID.
- If an environment already exists in TestOps and was previously used without explicit ID in report config, the existing environment ID from TestOps should be copied into config before renaming the environment.

```
environments: {
  8f3c2d1a4b5e6f7g8h9i0j1k2l3m4n5: {
    name: "QA EU",
    matcher: ...
  }
}
```


packages/cli/src/commands/qualityGate.ts
```diff
@@ -56,36 +63,32 @@ export class QualityGateCommand extends Command {
...
   cwd = Option.String("--cwd", {
     description: "The working directory for the command to run (default: current working directory)",
   });
 
   async execute() {
-    let normalizedEnvironment = this.environment;
-
-    if (typeof this.environment === "string") {
-      const envValidationResult = validateEnvironmentName(this.environment);
+    const environmentOptions = {
```
epszaw, line 75: In some places [in the PR we have a pair](https://github.com/allure-framework/allure3/pull/536/changes#diff-0ca230488e5d17c865ab7c627fc51d68bca22cd3f855ae8908a858b88713ba0bR121) of `environmentId` and `environment` properties, in another ones (like here) `environmentName` and `environment` properties – it's confusing, what `environment` property does mean, we can't create strong association.

Let's keep on CLI arguments level `--environment` and `--environment-name` options for backward compatibility (because `environment` actually didn't change it's purpose, but we just added approach to assign custom name to it). But in the code, let's use `environmentId` and `environmentName` properties pair. In this way, we would have a strong understanding without need of building associations and additional contexts around the properties.

epszaw (changes requested):

packages/cli/src/commands/qualityGate.ts
formaceft-93: fixed this one

packages/plugin-api/src/qualityGate.ts
```diff
@@ -6,6 +6,8 @@ export type QualityGateValidationResult = {
   actual: any;
   rule: string;
   message: string;
+  environmentId?: string;
```
epszaw: We can't extend the types, it's enough to keep `environment` property as is (the same also relates to `TestResult` one), it already allows us to get a specific environment's name.

Please, rollback these changes for `QualityGateValidationResult`, `TestResult` and similar things and use `environmentId`/`environmentName` only in variables and constants.

packages/plugin-api/src/config.ts
```diff
@@ -22,9 +22,13 @@ export interface Config {
...
   /**
-   * Environment which will be assigned to all tests
-   * Has higher priority than matched environment from the environments config field
+   * Environment display name which will be assigned to all tests.
+   */
+  environmentName?: string;
+  /**
+   * Legacy alias for environmentName.
    */
   environment?: string;
```
epszaw, line 30: It's not legacy, `environmentId` shouldn't be replacement for that. Keep the original option

packages/plugin-awesome/src/generateTimeline.ts
```diff
@@ -41,7 +45,8 @@ export const generateTimeline = async (writer: Writer, trs: AwesomeTestResult[],
       continue;
     }
 
-    const { host, thread } = test.groupedLabels;
+    const host = test.labels?.find(({ name }) => name === "host")?.value;
+    const thread = test.labels?.find(({ name }) => name === "thread")?.value;
```
epszaw, line 50: Check one more time, that the actual `test` (in the runtime) doesn't have `groupedLabels` property because it's simpler to use the property

packages/plugin-awesome/src/environments.ts
```diff
@@ -3,7 +3,7 @@ import type { TestResultFilter } from "@allurereport/plugin-api";
 
 export const filterEnv = (env: string, filter?: TestResultFilter) => {
   return (testResult: TestResult) => {
-    if (testResult.environment !== env) {
+    if ((testResult as TestResult & { environmentId?: string }).environmentId !== env) {
```
epszaw: The `TestResult` type has `environment` property. Use it instead

packages/core/src/report.ts
```diff
@@ -217,15 +226,17 @@ export class AllureReport {
...
   }) => {
-    const { trs, knownIssues, state, environment } = params;
+    const { trs, knownIssues, state, environmentId, environmentName } = params;
 
     return this.#qualityGate!.validate({
       trs: trs.filter(Boolean),
       knownIssues,
       state,
-      environment,
+      environmentId,
```
epszaw: It's enough to leave just `environment` as it was before because we store environment information separately, so we can get the data later using just environment id.

packages/core/src/report.ts
```diff
@@ -137,6 +144,8 @@ export class AllureReport {
       known,
       defaultLabels,
       environment,
+      environmentId,
+      environmentName,
```
epszaw: We already have `environment` and `environmentsConfig` properties here.
It's enough to use the properties to get `name` for any environment.

packages/core/src/report.ts
```diff
@@ -73,7 +73,12 @@ export class AllureReport {
   readonly reportUuid: string;
   reportUrl?: string;
 
-  constructor(opts: FullConfig) {
+  constructor(
+    opts: FullConfig & {
```
epszaw: Remove the type extension, see rest comments in the file to see the reason

packages/core-api/src/utils/environment.ts
```diff
@@ -55,17 +86,57 @@ export const assertValidEnvironmentName = (name: unknown, source: string = "envi
...
+
+  if (!match) {
+    return undefined;
+  }
+
+  const [id, descriptor] = match;
+
+  return [
+    id,
+    {
```
epszaw: Why can't we just return environment identity object here?

packages/core-api/src/utils/environment.ts
```diff
@@ -55,17 +86,57 @@ export const assertValidEnvironmentName = (name: unknown, source: string = "envi
...
+  ];
+};
+
+export const matchEnvironmentIdentity = (
+  envConfig: EnvironmentsConfig,
+  tr: Pick<TestResult, "labels">,
+): EnvironmentIdentity => {
+  const match = findMatchedEnvironmentEntry(envConfig, tr);
+
+  return match?.[1] ?? DEFAULT_ENVIRONMENT_IDENTITY;
```
epszaw: In case we changed `findMatchedEnvironmentEntry` signature, it's enough just to check `findMatchedEnvironmentEntry(envConfig, tr) ?? DEFAULT_ENVIRONMENT_IDENTITY`, or we can return `DEFAULT_ENVIRONMENT_IDENTITY` right in the `findMatchedEnvironmentEntry` function or directly in the usage place, so I don't see need to keep the `matchEnvironmentIdentity` function

packages/cli/src/utils/environment.ts
```diff
@@ -0,0 +1,141 @@
...
+  if (identityFromId && identityFromName && identityFromId.id !== identityFromName.id) {
+    throw new UsageError(
+      `${source}: environment id ${JSON.stringify(identityFromId.id)} and environment name ${JSON.stringify(identityFromName.name)} resolve to different environments`,
+    );
+  }
+
+  const identity = identityFromId ?? identityFromName ?? configIdentity;
+
+  return {
+    environmentId: identity?.id,
```
epszaw: Let's use `EnvironmentIdentity` type as a return one. Let's not to create new, almost similar entities.

packages/core/src/store/store.ts
```diff
@@ -179,16 +169,18 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
-  #environments: string[] = [];
+  #environments: EnvironmentIdentity[] = [];
 
   constructor(params?: {
     history?: AllureHistory;
     known?: KnownTestFailure[];
     realtimeDispatcher?: RealtimeEventsDispatcher;
     realtimeSubscriber?: RealtimeSubscriber;
     defaultLabels?: DefaultLabelsConfig;
+    environmentId?: string;
```
epszaw: See [this comment](https://github.com/allure-framework/allure3/pull/536/changes#r2973667414)

packages/plugin-api/src/config.ts
formaceft-93: fxd

packages/plugin-awesome/src/generateTimeline.ts
formaceft-93: fxd as well

packages/core-api/src/utils/environment.ts
```diff
@@ -55,17 +86,46 @@ export const assertValidEnvironmentName = (name: unknown, source: string = "envi
   return validationResult.normalized;
 };
 
+export const assertValidEnvironmentId = (environmentId: unknown, source: string = "environment id"): string => {
```
epszaw: The function isn't used anywhere except it tests. Remove the function

packages/core/test/qualityGate/qualityGate.test.ts
```diff
@@ -174,6 +174,39 @@ describe("QualityGate", () => {
       expect(fastFailed).toBe(false);
     });
 
+    it("should keep legacy shorthand environment ids in failed results", async () => {
```
epszaw: It's not legacy

packages/core/test/store/store.test.ts
```diff
@@ -1799,11 +1799,11 @@ describe("environments", () => {
           },
         }),
     ).toThrow(
-      'Invalid store constructor environmentsConfig["foo\\nbar"] "foo\\nbar": name must not contain control characters',
+      'store constructor environmentsConfig["foo\\nbar"]: id id must contain only latin letters, digits, underscores, and hyphens',
```
epszaw: It's pretty obvious where the error happened, the class name is listed in the trace. Also, there's double `id id`, remove the prefix

packages/core/test/store/store.test.ts
```diff
@@ -2705,6 +2729,372 @@ describe("dump state", () => {
     expect(allTestResults).toHaveLength(1);
   });
 
+  it("should restore legacy latest-attempt index shape and keep environment separation for new dumps", async () => {
```
epszaw: Why legacy? :)

packages/core/test/config.test.ts
```diff
@@ -418,7 +440,83 @@ describe("resolveConfig", () => {
...
+            matcher: () => true,
+          },
+        },
+      }),
+    ).rejects.toThrow(
+      'The provided Allure config contains invalid environments: config.environments[" QA "]: id id must contain only latin letters, digits, underscores, and hyphens',
+    );
+  });
+
+  it("should reject untrimmed environment ids instead of normalizing them", async () => {
```
epszaw: Trimming is pretty trivial, let's not reject values, but trim them and in case of post-trimming spaces – fail the validation

packages/core/src/store/store.ts
```diff
@@ -49,11 +50,28 @@ import type {
...
+  normalizeEnvironmentDescriptorMap,
+  resolveEnvironmentIdentity,
+  resolveStoredEnvironmentIdentity,
+} from "../utils/environment.js";
 import { isFlaky } from "../utils/flaky.js";
 import { getStatusTransition } from "../utils/new.js";
 import { testFixtureResultRawToState, testResultRawToState } from "./convert.js";
 
+type StoredTestResult = TestResult & {
+  environmentId?: string;
```
epszaw: Use just `environment`, no need to introduce new types for this

packages/core/src/store/store.ts
```diff
@@ -244,13 +234,24 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
 
     this.#addEnvironments(environments);
 
     this.#realtimeSubscriber?.onQualityGateResults(async (results: QualityGateValidationResult[]) => {
       this.#qualityGateResults.push(...results);
+      this.#addEnvironments(
+        results
+          .map((result) =>
+            this.#resolveStoredEnvironmentIdentity({
+              environment: (result as StoredQualityGateValidationResult).environmentId ?? result.environment,
```
epszaw: We already have `environment`, just use it

packages/plugin-testops/src/plugin.ts
```diff
@@ -61,9 +59,40 @@ export class TestopsPlugin implements Plugin {
     return this.#ci && this.#ci.type !== "local";
   }
 
+  async #environmentIdentitiesByTrId(store: AllureStore, trsToUpload: Array<{ id: string }>) {
```
epszaw: We don't need to store anything in the plugin, just use the store methods because we have direct access to it in the methods

packages/plugin-testops/src/plugin.ts
```diff
@@ -16,13 +21,17 @@ import { TestOpsClient } from "./client.js";
 import type { TestopsPluginOptions } from "./model.js";
 import { resolvePluginOptions, unwrapStepsAttachments } from "./utils.js";
 
+type UploadableTestResult = {
```
epszaw: 1. Just use `TestResult` and use `environment` property as an `id`
2. Pass map `{ envId: envName }` as a second argument to the upload function

packages/core/test/config.test.ts
```diff
@@ -366,15 +374,27 @@ describe("resolveConfig", () => {
...
     await expect(resolveConfig({ environment: "" })).rejects.toThrow(
-      "The provided Allure config contains invalid environment names: environment: name must not be empty",
+      "The provided Allure config contains invalid environments: environment: name must not be empty",
     );
   });
 
   it("should throw an error for invalid forced environment control characters", async () => {
     await expect(resolveConfig({ environment: "foo\nbar" })).rejects.toThrow(
-      "The provided Allure config contains invalid environment names: environment: name must not contain control characters",
+      "The provided Allure config contains invalid environments: environment: name must not contain control characters",
```
epszaw: `environments: environment: name` looks weird

packages/core/test/config.test.ts
```diff
@@ -418,7 +440,82 @@ describe("resolveConfig", () => {
...
+        },
+      },
+    });
+
+    expect(resolved.environment).toBe("qa_env");
+    expect(Object.keys(resolved.environments)).toEqual(["qa_env"]);
+    expect(resolved.environments?.qa_env?.name).toBe("QA Env");
+  });
+
+  it("should resolve configured environment display names through the public environment field", async () => {
```
epszaw: The test checks env id

packages/plugin-testops/src/client.ts
```diff
@@ -7,6 +14,11 @@ import pLimit from "p-limit";
 
 import type { TestOpsLaunch, TestOpsNamedEnv, TestOpsSession } from "./model.js";
 
+type UploadableTestResult = {
+  testResult: TestResult;
+  namedEnvironment?: EnvironmentIdentity;
```
epszaw: Why can't we use just use `TestResult` and `environment` property? Get the env name from the store before uploading and then take name by the id

packages/plugin-testops/src/client.ts
```diff
@@ -140,7 +152,7 @@ export class TestOpsClient {
     this.#session = data;
   }
 
-  async createNamedEnvs(environments: string[]) {
+  async createNamedEnvs(environments: Array<{ id: string; name: string }>) {
```
epszaw: Use `EnvironmentIdentity` here

packages/core/src/store/store.ts
```diff
@@ -976,25 +1122,64 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
       indexKnownByHistoryId = {},
       qualityGateResults = [],
     } = stateDump;
-    const normalizedEnvironments = environments.map((environmentName) =>
-      normalizeEnvironmentNameForStore(environmentName, {
-        source: "environment",
-        fallback: DEFAULT_ENVIRONMENT,
-      }),
-    );
+    const normalizedEnvironments = environments
```
epszaw, line 1243: Here we mostly rely on dump logic which creates report state dump with all the possible environments. Here and next you're iterating through all the test results, which is an expensive operation because we can have A HUGE amount of test results in the dump archive. 

So, it's better to add new environments from the dump, existing ones should be unchanged because environments' names are defined in the config and work for any source of data.

packages/core/src/utils/environment.ts
```diff
@@ -1,48 +1,315 @@
...
+  for (const [id, descriptor] of Object.entries(environmentsConfig)) {
+    if ((descriptor.name ?? id) === environmentName) {
+      return createIdentity(id, descriptor);
+    }
+  }
+
+  return undefined;
+};
+
+const normalizeStoredEnvironmentName = (environmentName: unknown): string | undefined => {
```
epszaw: Why can't we just use `validateEnvironmentName`? These wrappers almost do nothing, and we can just use original functions.

Review all these utilities and remove similar wrapper functions which just proxy the same result of each other.

packages/plugin-testops/src/plugin.ts
```diff
@@ -120,8 +156,14 @@ export class TestopsPlugin implements Plugin {
 
     trsProgressBar.render();
 
+    const uploadableTrs: UploadableTestResult[] = allTrsWithAttachments.map((tr) => {
```
epszaw: Avoid all the additional data iterations, it's costly operation for us

packages/core/src/store/store.ts
formaceft-93: fxd

PR #541, opened 2026-03-13
Allowed environments

formaceft-93:
Add allowedEnvironments support as a strict allowlist of environment IDs.
Grouped global attachments and global errors by environment.


epszaw (changes requested):
🔴 There are too many errors when I run `allure run` in sandbox when I added `foo` as allowed environment. The process should stop immediately when we have non-allowed environments in the configuration or CLI arguments
🔴 `default` environment should be allowed by default, now it throws an error when I add `foo` environment only, but shouldn't

epszaw (approved):

Issue #545, opened 2026-03-18
Attachments of generated report are duplicated for each plugin


valerii-lukomskyi:
**Describe the bug**
Test results attachments are duplicated across multiple plugins which increases the size of the report. 
2 plugins in config = x2 disk space usage, 4 plugins = x4 and so on.

**To Reproduce**
1. Configure allurerc with 2 awesome plugins:
```
reports: './store/reports',
history: './store/history.jsonl',
knownIssues: './store/known-issues.json',
plugins: {
  report1: {
    import: '@allurereport/plugin-awesome',
    options: {
      singleFile: false,
      reportLanguage: "en",
      reportName: "Report 1",
      groupBy: ["epic", "feature", "story"],
      filter: ({ labels }) => labels.find(({ name, value }) => name === "epic" && value === "Epic1")
    }
  },
  report2: {
    import: '@allurereport/plugin-awesome',
    options: {
      singleFile: false,
      reportLanguage: "en",
      reportName: "Report 2",
      groupBy: ["epic", "feature", "story"],
      filter: ({ labels }) => labels.find(({ name, value }) => name === "epic" && value === "Epic2")
    }
  }
}
```
2. Run tests to generate test-results
3. Generate allure report with command: `allure generate store/allure-results`
4. Check the test results for each report: report1: ~60 tests, report2: 2 tests.
5. Check the disk space usage of created reports:
```
➜  reports du -sh */ 2>/dev/null | sort -hr
220M	report1/
219M	report2
```
Each report has similar structure:
./store/reports/report1/data/attachments
./store/reports/report2/data/attachments

And the main disk space usage comes from these `attachments` folder, which I've figured out stores the same files.

**Expected behavior**
As I guess it's supposed to use shared attachments folders across all plugins.

**Additional context**
Here are the script I've run to compare data of 2 folders:
```
A="report1/data/attachments"
B="report2/data/attachments"

countA=$(ls "$A" | wc -l)
countB=$(ls "$B" | wc -l)
echo "Files in report1: $countA"
echo "Files in report2:        $countB"

# Find common filenames
comm -12 <(ls "$A" | sort) <(ls "$B" | sort) | wc -l | xargs echo "Common filenames:"

# Of the common files, check how many are identical (by md5)
identical=0
differ=0
while IFS= read -r f; do
  md5a=$(md5 -q "$A/$f")
  md5b=$(md5 -q "$B/$f")
  if [ "$md5a" = "$md5b" ]; then
    ((identical++))
[...]

Issue #486, opened 2026-02-10
Log files duplicated in generated report, exploding report size


jrett2:
**Describe the bug**
All the log files generated by a report are duplicated in every plugin

**To Reproduce**
1. copy https://github.com/allure-framework/allure3-demo/blob/main/allurerc.mjs
2. run 2 sets of tests, using the config to filter results between A and B
3. see that the log files for A and B are contained in the created report for A, and that the log files for A and B are contained in the created report for B

**Expected behavior**
The log files for a report are only for the results in that report.

**Screenshots**
Here you can see that project A has 0 tests, and project B has 1331 tests. Then you can see they have the same number of attachments:
```
jrett@jrett-macbookpro allure-report % cat A/summary.json
{"name":"proj A","stats":{"total":0},"status":"passed","duration":0,"createdAt":0,"plugin":"Awesome","newTests":[],"flakyTests":[],"retryTests":[],"meta":{"reportId":"6ba7b9b0-e2b0-46ac-b531-81002a07e882","singleFile":false,"withTestResultsLinks":true},"pullRequestHref":"","jobHref":"https://github.com/w/allure-report/actions/runs/21789745881"}%

jrett@jrett-macbookpro allure-report % cat B/summary.json
[...]

PR #547, opened 2026-03-18
reuse files when generate report within hardlinks(#545)

todti:
reduces disk usage and I/O when generating multiple reports into the same output directory by reusing identical files via hardlinks instead of writing duplicates.

- Adds content-hash based deduplication in FileSystemReportFiles with atomic tmp+rename and a safe fallback to regular writes when hardlinking isn’t possible.
- Adds unit tests and an e2e check ensuring shared static assets and attachments remain accessible across multiple awesome reports (inode check on non-Windows).

## Caveats / coverage notes

- The tricky case is cache invalidation: if a path that was used as the canonical file for some content hash is later rewritten with different content, the old hash must stop pointing at that path. The implementation now drops that stale canonical mapping before writing the new file, and the unit suite covers the `old content -> new content -> old content at another path` scenario.
- Writes and hardlink replacements go through a temporary file followed by `rename`, so readers should not observe a partially written target file.
- Hardlinking can fail on some filesystems or environments (for example cross-device output, permissions, or missing canonical source). Those errors are treated as recoverable and fall back to a regular write.
- Inode assertions are only meaningful on non-Windows platforms, so tests check inode reuse conditionally. On Windows the tests still validate that every report path remains readable and has the expected content.
- The e2e coverage focuses on the multi-awesome-report case: shared static assets and global attachments must stay accessible from both report directories while avoiding duplicate inodes where the platform supports it.

refer to #545 #486 

image


jrett2:
FYI this is a duplicate of https://github.com/allure-framework/allure3/issues/486

DineshKumarRA:
This would be really helpful for us, we are using multi plugins in allure report, around 10 different awesome plugins based on the test category which increasing the reporting size to around 2GB even tests are run in only 1 report and remaining all are 0

todti:
been chosen #743

PR #550, opened 2026-03-25
categories id

todti:
Added stable category IDs so category identity is no longer based on name.

How it works:
Each category rule can now define id + name.
During normalization, categories are matched/merged by id (not by name).
If id is missing, we keep backward compatibility by using name as fallback.
Awesome category nodes are generated from category.id, so renaming name does not create a new category.
Test results now carry both fields: ```categories: [{ id, name }]```.
Logic I introduce:
id: stable key.
name : mutable display label.
Validation + collision checks for normalized IDs to prevent ambiguous config.

```import { defineConfig } from "allure";
export default defineConfig({
  categories: {
    rules: [
      {
        id: "integration.product-errors", // stable ID 
        name: "Product Errors", // display name can be renamed
        matchers: { statuses: ["failed"] },
      },
      {
        id: "integration.test-errors",
        name: "Test Errors",
        matchers: { statuses: ["broken"] },
      },
    ],
  },
});```

### DIFF
--- a/packages/plugin-testops/src/uploadCategory.ts
+++ b/packages/plugin-testops/src/uploadCategory.ts
@@ -13,16 +13,12 @@ export const toUploadCategory = (
     const groupingFromTestResult = c.grouping?.length ? c.grouping : undefined;
     const contextCategory =
       groupingFromTestResult === undefined
-        ? contextCategories.find(
-            (category) => category.id === externalId || category.name === c.name,
-          )
+        ? contextCategories.find((category) => category.id === externalId || category.name === c.name)
         : undefined;
     return {
       externalId,
       name: c.name,
-      grouping:
-        groupingFromTestResult ??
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/plugin-testops/src/uploadCategory.ts [hunk 1]
```diff
@@ -1,92 +1,25 @@
...
-  if (!matched) {
-    return undefined;
-  }
-  return {
-    externalId: matched.name,
-    name: matched.name,
-    grouping: buildGrouping(tr, matched),
-  };
-};
+import { buildUploadCategoryGrouping, toUploadCategoryFromContext } from "./utils.js";
```
epszaw: I think, we can rename the file to `utils/categories.ts` and move all the related utils (such as `buildUploadCategoryGrouping`) into

PR #552, opened 2026-03-27
feat: update testops plugin

vdvukhzhilov:
Closes https://github.com/allure-framework/allure-org/issues/341

### DIFF
--- a/packages/plugin-testops/src/client.ts
+++ b/packages/plugin-testops/src/client.ts
@@ -1,8 +1,8 @@
-import { ClientRequest } from "http";
+import type { ClientRequest } from "http";
 
 import type { AttachmentLink, CiDescriptor, TestError, TestResult, TestStatus } from "@allurereport/core-api";
-import { QualityGateValidationResult } from "@allurereport/plugin-api";
-import axios, { AxiosError, AxiosInstance } from "axios";
+import type { QualityGateValidationResult } from "@allurereport/plugin-api";
+import axios, { AxiosError, isAxiosError, type AxiosInstance } from "axios";
 import FormData from "form-data";
 import { chunk } from "lodash-es";
 import pLimit from "p-limit";
@@ -12,7 +12,6 @@ import { Logger } from "./logger.js";
 import type {
   AttachmentForUpload,
   AttachmentsResolver,
-  TestOpsPluginTestResult,
   FixtureResolver,
   LaunchCategoryBulkItem,
   LaunchCategoryBulkResult,
@@ -20,6 +19,7 @@ import type {
   TestOpsLaunch,
   TestOpsLaunchQualityGate,
   TestOpsNamedEnv,
+  TestOpsPluginTestResult,
   TestOpsSession,
 } from "./model.js";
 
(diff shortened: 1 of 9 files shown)


packages/plugin-testops/test/scenario.test.ts
```diff
@@ -0,0 +1,311 @@
+import { TestResult } from "@allurereport/core-api";
```
epszaw: Rename the file to `qualityGate.test.ts`

packages/plugin-testops/src/client.ts
```diff
@@ -83,6 +90,14 @@ export class TestOpsClient {
     }
   }
 
+  isTestOpsClientError(error: unknown): error is TestOpsClientError {
+    return (
+      error instanceof AxiosError &&
```
epszaw: Use `isAxiosError` util, it contains additional predicates

packages/plugin-testops/src/client.ts
```diff
@@ -291,87 +365,146 @@ export class TestOpsClient {
...
-          });
+          const extendedTestResult: TestOpsPluginTestResult = {
+            ...testResult,
+            // pass the report id to TestOps to be able to match the test result with the report
+            uuid: testResult.id,
+          };
+
+          const namedEnvironment =
+            !!testResult.environment &&
+            // @TODO: USE IDS HERE
```
epszaw: We treat `environment` property as env id, not a name, so some `TODO` comments are irrelevant

packages/plugin-testops/src/plugin.ts
```diff
@@ -50,55 +60,171 @@ export class TestopsPlugin implements Plugin {
...
+      progressBar.terminate();
+
+      if (error instanceof Error) {
+        if (this.#client.isTestOpsClientError(error)) {
+          this.#logger.error(`Failed to upload named environments: ${error.response.data.message}`);
+        } else {
+          this.#logger.error(`Failed to upload named environments: ${error.message}`);
+        }
+
+        if (isAxiosError(error)) {
```
epszaw: We don't perform http request anywhere except testops, so I'm not sure about the check. Guess `isTestOpsClientError` is enough to classify http error and handle rest errors in a simpler way (whatever just log or throw them)

packages/plugin-testops/src/plugin.ts
```diff
@@ -50,55 +60,171 @@ export class TestopsPlugin implements Plugin {
...
+      progressBar.update(0);
+      await this.#client.uploadGlobalAttachments({
+        attachments,
         attachmentsResolver: async (attachmentLink) => {
           const content = await store.attachmentContentById(attachmentLink.id);
-          const body = content ? await content.readContent(async (stream) => stream) : undefined;
-          // @ts-expect-error
-          const attachmentName = attachmentLink.name || attachmentLink.originalFileName;
+          const body = await content?.readContent(async (stream) => stream);
+          // @ts-expect-error - don't know
```
epszaw: Definitely, we need to mark it as `FIXME` because we receive `name` property in the attachment object in runtime, but the type doesn't contain the property at all and using `originalFileName` only make TestOps displaying wrong attachment name

packages/plugin-testops/src/logger.ts
```diff
@@ -0,0 +1,159 @@
...
+function getLogLevelFromEnv(): LogLevel {
+  if (env.LOG_LEVEL) {
+    return env.LOG_LEVEL as LogLevel;
+  }
+
+  if (env.ALLURE_LOG_LEVEL) {
+    return env.ALLURE_LOG_LEVEL as LogLevel;
+  }
+
+  if (env.CI) {
```
epszaw: We can remove the branch because we have `info` level by default

packages/plugin-testops/src/plugin.ts
vdvukhzhilov: Well, there can be some errors on the way to testops server, like network stuff, not responding cloudflare, etc

PR #578, opened 2026-04-13
Security: Weak Cryptographic Hash Function Used

ed-thuando:
## Summary

Security: Weak Cryptographic Hash Function Used

## Problem

**Severity**: `High` | **File**: `packages/plugin-api/src/utils/misc.ts:L1`

The MD5 hash function is used in packages/plugin-api/src/utils/misc.ts for creating message digests. MD5 is cryptographically broken and vulnerable to collision attacks. While the specific usage context determines actual risk, using MD5 for security-related purposes is dangerous and should be replaced with SHA-256 or another modern cryptographic hash function.

## Solution

Replace MD5 with SHA-256 or another secure hash function. Use node:crypto's createHash('sha256') instead of createHash('md5').

## Changes

- `packages/plugin-api/src/utils/misc.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

---
*Generated by [ContribAI](https://github.com/tang-vu/ContribAI) v6.0.0*

### DIFF
--- a/packages/plugin-api/src/utils/misc.ts
+++ b/packages/plugin-api/src/utils/misc.ts
@@ -1,3 +1,3 @@
 import { createHash } from "node:crypto";
 
-export const md5 = (data: string) => createHash("md5").update(data).digest("hex");
+export const sha256 = (data: string) => createHash("sha256").update(data).digest("hex");


baev:
not all hash functions are used for cryptographic reasons

PR #611, opened 2026-04-28
New service api

epszaw:
New service API compatible with https://github.com/allure-framework/allure-report-storage


packages/core/src/report.ts
```diff
@@ -293,14 +312,14 @@ export class AllureReport {
...
     // create remote report to publish files into
-    if (this.#allureServiceClient && this.#publish && branch) {
-      const { url } = await this.#allureServiceClient.createReport({
+    if (this.#allureServiceClient && this.#publish) {
+      const url = await this.#allureServiceClient.createReport({
         reportUuid: this.reportUuid,
         reportName: this.#reportName,
-        branch,
+        ...remoteParams,
       });
```
vdvukhzhilov, line 320: Before this would not run if there is no `branch` passed.
Now it seems to run if we pass neither `branch` nor `repo`.
Is that okay? 🤔

vdvukhzhilov (approved):

packages/core/src/report.ts
epszaw: Currently it's ok because we plan to implement more clients, which aren't going to utilize branch parameters at all.

The legacy client also doesn't use the `repo` parameter.

vdvukhzhilov: 👌

PR #619, opened 2026-05-04
Redo retry handling and remove legacy retry markers

vdvukhzhilov:
# Summary

This PR updates how retries are identified and handled in the report pipeline.

The logic is now more explicit and consistent across the project, and retry-related naming is clearer in both data models and APIs.
As part of this cleanup, the old retry markers were removed and replaced with the new approach.


packages/core/src/store/retrySubstore.ts
```diff
@@ -0,0 +1,85 @@
...
+    return undefined;
+  }
+
+  return md5(`${testCaseId}:${parametersHash}:${environmentId ?? "default"}`);
+};
+
+export class RetrySubstore {
+  readonly #testResultsByRetryHash: Map<string, TestResult[]> = new Map();
+
+  constructor(private readonly store: AllureStore) {}
```
delatrie: nit: `store` is unused, consider removing

packages/core/test/store/store.test.ts
```diff
@@ -391,6 +391,72 @@ describe("test results", () => {
...
+    );
+
+    const [latest] = await store.allTestResults();
+    const retries = await store.retriesByTrId(latest.id);
+
+    expect(latest.retryHash).toBeDefined();
+    expect(retries.map(({ name }) => name)).toEqual(["retry"]);
+  });
+
+  it("should respect provided parametersHash when building retryHash", async () => {
```
delatrie, line 544: nit: For completeness, one more test can be added: same parameters + different `parameterHash`s should give unrelated test results (not retries).

packages/core/test/store/store.test.ts
```diff
@@ -3060,44 +3124,55 @@ describe("dump state", () => {
...
-    const latestAttemptsByEnv = store.dumpState().indexLatestEnvTestResultByHistoryId;
+    const dump = store.dumpState();
 
-    expect(Object.keys(latestAttemptsByEnv)).toEqual(["qa"]);
-    expect(Object.keys(latestAttemptsByEnv.qa)).toHaveLength(1);
-    expect(Object.values(latestAttemptsByEnv.qa)).toEqual([expect.any(String)]);
+    expect(dump.indexLatestEnvTestResultByHistoryId).toEqual({});
+    expect(dump.testResults[Object.keys(dump.testResults)[0]]).toEqual(
+      expect.objectContaining({
+        retryHash: expect.any(String),
```
delatrie: The test has changed the semantics.

Before: check if the dump associates an environment with the latest result of each test.

After: check if the dump contains the test result's retryHash.

IMO, we should keep both, as they check slightly different aspects of dump.

Perhaps a better way to verify that the latest attempts are assigned to correct environments, given the changes to dump internals, is to run `restoreState` and check the test results.

If this is troublesome, I suggest at least renaming the test.

packages/core/src/store/retrySubstore.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const stringifyRetryParameters = (parameters: TestParameter[] = []): string =>
+  parameters
+    .filter((parameter) => !parameter.excluded)
+    .sort(compareByNameThenValue)
+    .map((parameter) => `${parameter.name}:${parameter.value}`)
+    .join(",");
+
+const compareByStartAsc = (first: TestResult, second: TestResult) => {
+  if (first.start === undefined || second.start === undefined || first.start === second.start) {
```
delatrie: This is an ill-formed comparer: for a = 1, b = undefined, and c = 2, it gives a = b, b = c, but a < b. In other words, it's not transitive. Also, it doesn't account for `null`.

packages/core/src/store/retrySubstore.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    const attempts = this.#testResultsByRetryHash.get(testResult.retryHash) ?? [];
+    const index = attempts.findIndex((attempt) => attempt.id === testResult.id);
+
+    if (index <= 0) {
+      return NO_RETRIES;
+    }
+
+    return attempts.slice(0, index).sort(nullsLast(compareBy("start", reverse(ordinal()))));
```
delatrie: There is no point in sorting twice. We should either sort in `upsert` or in `retriesByTr`.

If you choose sorting in `upsert`, make sure the order matches `nullsLast(compareBy("start", reverse(ordinal())))` (currently, is does not).

packages/core/src/store/store.ts
```diff
@@ -76,37 +74,58 @@ const index = <T>(indexMap: Map<string, T[]>, key: string | undefined, ...items:
...
-  first.start === undefined || second.start === undefined || first.start < second.start;
+const fallbackTestCaseIdLabelName = "_fallbackTestCaseId";
 
-const hidePreviousAttemptByEnvironment = (
-  state: Map<string, Map<string, TestResult>>,
-  testResult: TestResult,
-  environmentId: string | undefined,
-) => {
-  const { historyId } = testResult;
+const getFallbackHistoryId = (testResult: Pick<TestResult, "labels" | "parameters">): string | undefined => {
```
delatrie: Use `findLast` instead of `[...] + reverse` + `find`

packages/core/src/store/store.ts
```diff
@@ -76,37 +74,58 @@ const index = <T>(indexMap: Map<string, T[]>, key: string | undefined, ...items:
...
-      } else {
-        historyIdToLastAttemptResult.set(historyId, testResult);
+const getHistoryIdCandidates = (testResult: Pick<TestResult, "historyId" | "labels" | "parameters">): string[] => {
+  const result: string[] = [];
+
+  if (testResult.historyId) {
+    result.push(testResult.historyId);
+  }
+
+  const fallbackHistoryId = getFallbackHistoryId(testResult);
```
delatrie: `getFallbackHistoryId` uses the latest `_fallbackTestCaseId`. Should we also consider earlier values? Or have we already decided to use only the last one?

delatrie (changes requested):

packages/core/src/store/retrySubstore.ts
vdvukhzhilov: Yeap, there was an idea to get something from it, but I deleted the code and left this :(

packages/core/test/store/store.test.ts
vdvukhzhilov: As I remember the logic "the dump associates an environment with the latest result of each test" was somehow connected to calculating retries and after refactor got redundant

I'll check closer

packages/core/src/store/retrySubstore.ts
vdvukhzhilov: But can start be `null` though? 🤔 

But good catch anyway! 👍

packages/core/src/store/store.ts
vdvukhzhilov: Deleted this thing entirely 👍

packages/core/src/store/retrySubstore.ts
vdvukhzhilov: Yeah, I store them as earliest start - latest start, that's why I sort them again here.
I'll make it simple 👌

packages/core/test/store/store.test.ts
vdvukhzhilov: done

packages/core/src/store/retrySubstore.ts
```diff
@@ -0,0 +1,121 @@
...
+        continue;
+      }
+
+      this.#testResultIngestIndexById.set(id, this.#testResultIngestIndexById.size);
+    }
+  }
+
+  ingestOrderIdsForDump(): string[] {
+    return [...this.#testResultIngestIndexById.entries()]
+      .sort(([, indexA], [, indexB]) => indexA - indexB)
```
delatrie: nit: maps in JS guarantee the insertion order, which means indices are already in increasing order

packages/core/src/store/retrySubstore.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  #getIngestIndex(testResultId: string): number {
+    return this.#testResultIngestIndexById.get(testResultId) ?? -1;
+  }
+
+  #compareResults(first: TestResult, second: TestResult): -1 | 0 | 1 {
+    const indexA = this.#getIngestIndex(first.id);
+    const indexB = this.#getIngestIndex(second.id);
+
+    if (typeof first.start === "number" && typeof second.start === "number" && first.start !== second.start) {
```
delatrie: If only one item has `start`, that item should always come before (or always after) the other. Otherwise, if `A=1, B=undefined, C=0` (inserted from left to right), the algorithm may give the `A, B, C` order, since from the comparison function standpoint, `A` and `B` are already in order, as well as `B` and `C`.

In reality, this will rarely happen. The only case I can think of is when there are multiple sources of test results (e.g., allure2 + xcresult), and some tests accidentally got the same testCaseId. Possible but unlikely, but worth fixing, I guess.

delatrie (changes requested):

packages/core/src/store/retrySubstore.ts
vdvukhzhilov: Ah, yes, valid point

packages/core/src/store/retrySubstore.ts
vdvukhzhilov: thanks, I still think of `Map` as a synthetic sugar that is polyfill'ed in final code😄

delatrie (approved):

PR #622, opened 2026-05-05
fix sorting

formaceft-93:
Fix sorting problems

### DIFF
--- a/packages/web-awesome/src/utils/treeFilters.ts
+++ b/packages/web-awesome/src/utils/treeFilters.ts
@@ -58,18 +58,27 @@ const groupComparatorByTreeSortBy = (sortBy: SortBy = "status,asc"): Comparator<
   }
 };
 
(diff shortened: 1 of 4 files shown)


packages/web-awesome/src/utils/treeFilters.ts
```diff
@@ -58,18 +58,27 @@ const groupComparatorByTreeSortBy = (sortBy: SortBy = "status,asc"): Comparator<
   }
 };
 
+const withDirection = <T extends { name: string }>(cmp: Comparator<T>, sortBy: SortBy): Comparator<T> => {
+  const isStatusSort = sortBy.startsWith("status,");
```
epszaw: In my opinion it's not so safe to check sort predicates in that way; it's much safer to use the previous approach with `split(",")` and check parameters' existence by `.includes("status")` or `.include("desc")`

epszaw (changes requested):

epszaw (approved):

PR #625, opened 2026-05-06
Allure check command

epszaw:
## Summary

Adds a new `allure check` command for recording external check results.

The command supports two modes:

- command mode: `allure check --name "Lint" -- npm run lint`
- manual mode: `allure check --name "Manual approval" --status passed --message "approved" --tag release`

Check results are written as `<uuid>-check-result.json` files into the configured Allure output directory, or stored in the state dump when `--dump` is provided.

## Changes

- Added `AllureCheckResult` / `AllureCheckStatus` to `core-api`.
- Added `allure check` CLI command with:
  - exit-code based `passed` / `failed` status mapping
  - stdout capture into `details.message`
  - stderr capture into `details.error`
  - full command string capture into `details.command`
  - repeated `--tag` support
  - `--output`, `--cwd`, `--config`, and `--dump` support
- Ensured check command execution uses the resolved current working directory.
- Added store API support for check results:
  - `addCheckResult`
  - `allCheckResults`
- Added check result dump/restore support via `check-results.json`.
- Added Allure 2 reader support for `*-check-result.json`.
- Added pluginless config override behavior using `plugins: {}` so `allure check` can initialize report state without running configured plugins.

## Testing

- `yarn workspace @allurereport/core build`
- `yarn workspace allure build`
- `yarn allure agent ... -- yarn workspace @allurereport/core test test/config.test.ts`
- `yarn allure agent ... -- yarn workspace allure test test/commands/check.test.ts`
- `git diff --check main`


packages/cli/src/commands/check.ts
```diff
@@ -0,0 +1,227 @@
...
+  throw new UsageError(`Invalid --status value ${JSON.stringify(status)}. Expected one of: passed, failed`);
+};
+
+type CheckCommandOutput = {
+  stdout: string;
+  stderr: string;
+};
+
+const normalizeOutput = (value: string | undefined) => {
+  const trimmed = value?.trimEnd();
```
formaceft-93: Why only end needs trim? (like  normalizeOutput('   hello    ') for example )

packages/cli/src/commands/check.ts
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    const message = normalizeOutput(commandOutput ? commandOutput.stdout : this.message);
+    const error = normalizeOutput(commandOutput?.stderr);
+    const result = buildCheckResult(this.name, status, this.tag, checkCommand?.commandLine ?? "", message, error);
+    const config = await readConfig(cwd, this.config, {
+      output: this.output,
+      plugins: {},
+    });
+    const checkConfig = {
```
formaceft-93, line 205: readConfig should already return "plugins: []", so do we need this override?

formaceft-93 (approved):

packages/cli/src/commands/check.ts
```diff
@@ -0,0 +1,227 @@
...
+    details: "This command records an Allure check result based on a command exit code or an explicit status.",
+    examples: [
+      ['check --name "Lint" -- npm run lint', "Run npm run lint and record the check result"],
+      ['check --name "Manual approval" --status passed --tag release', "Record a manual passed check result"],
+      ['check --name "Lint" --dump=checks -- npm run lint', "Store the check result in checks.zip instead of output"],
+    ],
+  });
+
+  name = Option.String("--name", {
+    description: "The check name",
```
formaceft-93, line 129: We can add "required: true" as well and then I think we won't need to check name here https://github.com/allure-framework/allure3/pull/625/changes#diff-3ec6d4fa0ab6937cd35906dc5435df927e86e0be3012ed1ca85c93bb2757c829R172

packages/cli/src/commands/check.ts
epszaw: 👍

packages/cli/src/commands/check.ts
epszaw: No, it shouldn't. By default it returns config with awesome and agent plugins

packages/cli/src/commands/check.ts
epszaw: 👍

formaceft-93 (approved):

PR #633, opened 2026-05-08
Optimize allure3

formaceft-93:
### Frontend 
- **Lazy sections** (`web-awesome`): Charts and Timeline are loaded on demand via dynamic imports. A component cache makes tab switching instant, and if a chunk fails to load we fall back to Report instead of crashing.

### Generator

- **Bulk store reads** (`plugin-api`, `plugin-awesome`): `generateTestResults` now fetches fixtures, history, retries, and attachments in a single `relatedByTestResultIds` call instead of querying per test result. This is the main reason the 50k dataset generator drops from ~69s to ~29s.

### Reader

- **Early file matching** (`reader`): the `allure2` reader uses a `matches()` pattern to skip irrelevant files immediately instead of trying to parse everything.

### Allure Awesome — before vs after

| Dataset | Gen before | Gen after | Gen Δ | Report before | Report after | Report Δ |
|---|---|---|---|---|---|---|
| 10k | 11.43s | 8.04s | 1.42x | 9.12s | 5.38s | 1.69x |
| 50k | 68.85s | 28.97s | 2.38x | 66.51s | 26.33s | 2.53x |

| Dataset | Load before | Load after | Speedup | Heap before | Heap after |
|---|---|---|---|---|---|
| 10k | 2216ms | 1351ms | 1.64x | 272.8 MB | 23.4 MB |
| 50k | 6939ms | 1381ms | 5.03x | 631.3 MB | 68.9 MB |

### DIFF
--- a/packages/plugin-api/src/store.ts
+++ b/packages/plugin-api/src/store.ts
@@ -65,7 +65,7 @@ export interface AllureStore {
   retriesByTrId: (trId: string) => Promise<TestResult[]>;
   historyByTrId: (trId: string) => Promise<HistoryTestResult[] | undefined>;
   fixturesByTrId: (trId: string) => Promise<TestFixtureResult[]>;
-  relatedByTestResultIds?: (trIds: readonly string[]) => Promise<TestResultRelatedData>;
+  relatedByTestResultIds: (trIds: readonly string[]) => Promise<TestResultRelatedData>;
   // aggregate api
   failedTestResults: () => Promise<TestResult[]>;
   unknownFailedTestResults: () => Promise<TestResult[]>;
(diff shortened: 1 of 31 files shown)


packages/plugin-api/src/store.ts
```diff
@@ -78,6 +86,43 @@ export interface AllureStore {
   envVariablesByEnvironmentId: (envId: string) => Promise<Record<string, any>>;
 }
 
+export const relatedByTestResultIds = async (
```
epszaw: Move to the store itself. In `@allurereport/plugin-api` we keep types mostly. Let's hand the convention

epszaw (changes requested):

epszaw (approved):

PR #653, opened 2026-05-15
Add multi-field search

formaceft-93:
Adds a dedicated search payload for Awesome reports and switches the tree query search to MiniSearch. Search now works beyond test name/id and can match fields like fullName, historyId, owner and visible labels, categories, top-level error messages, and links.

The search data is written separately as widgets/search-index.json and per environment as widgets/<env>/search-index.json, so tree.json stays focused on rendering the tree and multi-environment reports only search within the selected environment. The frontend loads and builds the MiniSearch index lazily when a query is entered, then intersects search matches with existing status/tag/category filters.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214697087944551


packages/web-awesome/src/stores/search.ts
```diff
@@ -0,0 +1,139 @@
...
+    searchOptions: {
+      combineWith: "AND",
+      prefix: true,
+      fuzzy: (term) => (term.length > 3 ? 0.2 : false),
+      maxFuzzy: 2,
+      boost: {
+        name: 4,
+        fullName: 3,
+        owner: 3,
+        statusMessage: 2,
```
baev: this one should be used last, I would say the order is 

 "id",
  "name",
  "fullName",
  "owner",
  "labels",
  "links",
  "categories",
  "statusMessage",
  "historyId",

baev (changes requested):

baev (approved):

PR #657, opened 2026-05-20
fix playwright trace

### LINKED ISSUE #624: unable to open playwright trace file natively in allure3 report
**Describe the bug**
playwright trace file cannot be opened natively within allure3 report anymore
**To Reproduce**
1. run a playwright test with failure to generate a trace file
2. generate allure3 report - npx allure generate ...
3. open report - npx allure open
4. click to open trace file
**Expected behavior**
trace file open natively

**Additional context**
It looks like current approach is to embed trace.playwright.dev like an iframe? - [HERE](https://github.com/allure-framework/allure3/blob/990dd740cdeb240d3d600c7285b6bb60721b9009/packages/web-awesome/src/components/TestResult/TrPwTraces/openPwTraceInNewTab.ts#L17). If so, some recent playwright repo change might be blocking it - [HERE](https://github.com/microsoft/playwright/pull/40548/changes)
If the above is true/confirmed...Do we have path/solution to address it?

formaceft-93:
Playwright Trace Viewer changed its cross-origin `postMessage` handling in microsoft/playwright#40548, which broke the old Allure flow for opening traces.

This PR switches to the official `trace.playwright.dev/?trace=<url>` API and adds CORS for trace attachment requests so the viewer can fetch the file directly.

Single-file reports are handled with a manual fallback because their attachments are embedded as `data:` URLs and cannot be passed to the external viewer as a regular file URL.

Fixes allure-framework/allure3#624

### DIFF
--- a/packages/e2e/test/allure-awesome/features/attachments.test.ts
+++ b/packages/e2e/test/allure-awesome/features/attachments.test.ts
@@ -472,14 +472,15 @@ test.describe("attachments", () => {
 
       await testResultPage.testResultAttachmentLocator
         .filter({ has: page.getByText("trace", { exact: true }) })
-        .getByRole("button")
-        .nth(0)
+        .first()
+        .getByLabel("Open Playwright Trace")
         .click();
 
       const popup = await popupPromise;
 
       if (popup) {
-        await popup.waitForURL("https://trace.playwright.dev/", { timeout: 5_000 });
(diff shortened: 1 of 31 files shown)


epszaw (approved):

baev:
the change is reverted by PW, so no fix is needed at the moment

PR #658, opened 2026-05-20
feat(plugin-testops): attach git information on launch create

vdvukhzhilov:
## Summary

When an Allure report is uploaded to TestOps from CI, the launch can now carry git context — repo, commit, branch, and PR when the pipeline provides it

Collection runs only in CI and only if turned it on

### DIFF
--- a/packages/ci/src/helpers/github.ts
+++ b/packages/ci/src/helpers/github.ts
@@ -2,19 +2,29 @@ import { readFileSync } from "node:fs";
 
 import { getEnv } from "../utils.js";
 
-const pullRequestRefNameRe = /^(\d+)\/merge$/;
-const pullRequestMergeRefRe = /^refs\/pull\/(\d+)\/merge$/;
+const PULL_REQUEST_REF_NAME_RE = /^(\d+)\/merge$/;
+const PULL_REQUEST_MERGE_REF_RE = /^refs\/pull\/(\d+)\/merge$/;
+
+type GithubPullRequestEvent = {
+  number?: number | string;
+  pull_request?: {
+    number?: number | string;
+  };
+};
+
+const normalizePullRequestNumber = (value: number | string | undefined): string => {
+  if (value === undefined || value === null) {
+    return "";
+  }
+
+  return String(value);
+};
 
 export const parsePullRequestNumberFromEventJson = (content: string): string => {
   try {
-    const event = JSON.parse(content) as { pull_request?: { number?: number } };
-    const number = event?.pull_request?.number;
+    const event = JSON.parse(content) as GithubPullRequestEvent;
 
-    if (number === undefined || number === null) {
-      return "";
-    }
-
-    return String(number);
+    return normalizePullRequestNumber(event?.number ?? event?.pull_request?.number);
   } catch {
     return "";
   }
@@ -38,14 +48,14 @@ const readPullRequestNumberFromEventPath = (): string => {
 
 export const resolveGithubPullRequestNumber = (): string => {
   const refName = getEnv("GITHUB_REF_NAME") || "";
-  const refNameMatch = refName.match(pullRequestRefNameRe);
+  const refNameMatch = refName.match(PULL_REQUEST_REF_NAME_RE);
 
   if (refNameMatch) {
     return refNameMatch[1];
   }
 
   const githubRef = getEnv("GITHUB_REF") || "";
-  const mergeRefMatch = githubRef.match(pullRequestMergeRefRe);
+  const mergeRefMatch = githubRef.match(PULL_REQUEST_MERGE_REF_RE);
 
   if (mergeRefMatch) {
     return mergeRefMatch[1];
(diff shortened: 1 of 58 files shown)


packages/ci/src/gitHints/amazon.ts
```diff
@@ -0,0 +1,72 @@
+import { type CiDescriptor, type CiGitHints } from "@allurereport/core-api";
```
epszaw: Let's move the helpers/hints to related detectors. I think current solution a bit overengineered, when the detector designed to provide all the required information in a single place

packages/ci/src/gitHints/amazon.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const headRef = getEnv("CODEBUILD_WEBHOOK_HEAD_REF");
+
+  return {
+    provider: repository.provider,
+    repository: {
+      slug: repository.slug,
+      url: repository.url,
+    },
+    sourceBranch:
```
epszaw: The branch regex stops at the first `/`, so names like `feature/foo` get truncated to feature. Fix the issues and add appropriate tests to validate it.

packages/ci/src/helpers/github.ts
```diff
@@ -0,0 +1,61 @@
+import { readFileSync } from "node:fs";
+
+import { getEnv } from "../utils.js";
+
+const pullRequestSuffixRe = /\/merge$/;
```
epszaw: Let's use UPPER_SNAKECASE

packages/ci/src/helpers/github.ts
```diff
@@ -0,0 +1,61 @@
...
+const pullRequestMergeRefRe = /^refs\/pull\/(\d+)\/merge$/;
```
epszaw: Any GITHUB_REF_NAME ending in `/merge` is treated as a PR, which can misclassify normal branches like `release/merge`.

epszaw (changes requested):

packages/plugin-testops/src/gitFlow/options.ts
```diff
@@ -0,0 +1,62 @@
+import { env } from "node:process";
+
+import { DEFAULT_ANCESTOR_LIMIT } from "@allurereport/git";
+
+import type { TestOpsPluginOptions } from "../model.js";
+export type ResolvedGitFlowOptions = {
```
epszaw: Move to the `types`

packages/ci/src/gitHints/amazon.ts
epszaw: Nothing has changed

epszaw (changes requested):

packages/ci/src/helpers/github.ts
vdvukhzhilov: now only GITHUB_REF checked

packages/git/src/collectGitFacts.ts
```diff
@@ -0,0 +1,61 @@
...
+const collectLocalState = (cwd?: string): GitLocalState => {
+  const statusPorcelain = runGit(["status", "--porcelain"], cwd) ?? "";
+  const branchRaw = runGit(["rev-parse", "--abbrev-ref", "HEAD"], cwd);
+  const detachedHead = branchRaw === "HEAD";
+  const upstreamCommit = runGit(["rev-parse", "--verify", "@{u}"], cwd);
+  const upstreamBranch = runGit(["rev-parse", "--abbrev-ref", "@{u}"], cwd);
+
+  return {
+    uncommittedChanges: statusPorcelain.length > 0,
+    unpublishedCommit: !upstreamCommit,
```
delatrie: This returns `false` for:

```
local:   HEAD(branch)  -> C2 -> C1
remote:  remote/branch -> C1
```

While it should return `true` because local `C2` is unpublished.

The real check should be something like `[git rev-parse --verify "HEAD"] != [git rev-parse --verify "@{u}"]`

packages/git/src/collectGitFacts.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  if (!revList) {
+    return undefined;
+  }
+
+  const firstParentAncestors = revList
+    .split("\n")
+    .map((line) => line.trim())
+    .filter(Boolean)
+    .filter((hash) => hash !== commit);
```
delatrie: nit: HEAD is always the first one, so `.slice(1)` should do the job.

packages/git/src/collectGitFacts.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  const commit = runGit(["rev-parse", "HEAD"], cwd);
+
+  if (!commit) {
+    return undefined;
+  }
+
+  const branchRaw = runGit(["rev-parse", "--abbrev-ref", "HEAD"], cwd);
+  const branch = branchRaw && branchRaw !== "HEAD" ? branchRaw : undefined;
```
delatrie: Shall we prefer the remote branch name if available (with the remote prefix stripped)? It's possible the local branch name doesn't match the remote one.

delatrie (changes requested):

delatrie (approved):

epszaw (approved):

PR #671, opened 2026-05-28
Unified service client

epszaw:


packages/service/src/model.ts
```diff
@@ -11,10 +11,35 @@ export const ALLURE_LOGIN_EXCHANGE_TOKEN_PATH = join(ALLURE_FILES_DIRNAME, "exch
 
 export const ALLURE_ACCESS_TOKEN_PATH = join(ALLURE_FILES_DIRNAME, "access_token");
 
+export const ALLURE_SERVICE_STORAGE_PREFIX = "ars1.";
+
+export const ALLURE_SERIVCE_TESTOPS_PREFIX = "ato1.";
```
todti: typo

packages/service/src/utils/http.ts
```diff
@@ -84,3 +171,143 @@ export const createServiceHttpClient = (serviceUrl: string, accessToken: string)
...
+
+      failedUploads.add(filename);
+
+      if (failedUploads.size > maxSimultaneousFailures || attempt >= maxAttempts) {
+        throw error;
+      }
+    }
+  }
+
+  return false;
```
todti, line 207: looks unreachable, thought `throw error` above will do safety return

packages/core/src/config.ts
```diff
@@ -342,7 +352,27 @@ export const resolveConfig = async (config: Config, override: ConfigOverride = {
     plugins: pluginInstances,
     defaultLabels: config.defaultLabels ?? {},
     qualityGate: config.qualityGate,
-    allureService: config.allureService,
+    allureService: config.allureService
+      ? {
+          accessToken: config.allureService?.accessToken,
```
todti: `config.allureService ?` will guarantee non-nullish, maybe just keep `.` instead of `?`?

todti (approved):

packages/service/src/model.ts
epszaw: 👍

packages/service/src/utils/http.ts
epszaw: It's the typescript type placeholder only

packages/core/src/config.ts
epszaw: Sure

Issue #431, opened 2026-01-05
Expanded html attachment appears in raw format (awesome)


Oksana-Parysheva:
**Describe the bug**
Html attachment file is displayed in raw format when expanded inside test details. When click on Expand for it - it shows formatted.

**To Reproduce**
1. Go to any test case result
2. Click on test step
3. Scroll down to html attachment and click on it to expand

**Expected behavior**
It is expected to see formatted html text (in example is a table) without clicking Expand arrows)

**Screenshots**
<img width="812" height="360" alt="Image" src="https://github.com/user-attachments/assets/40a16e35-60a8-473c-b8e6-db135cc2f0d3" />

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome v.143.0.7499.41

**Additional context**

kieran-ryan:
Closed by #697 and #669.

svasenkov:
Looks like this was already fixed in #697 / #669 (per @kieran-ryan). Could a maintainer please close this issue if nothing remains?

epszaw:
Hi! This appears to be resolved now. The HTML attachment rendering was fixed in [#697](https://github.com/allure-framework/allure3/pull/697) and [#669](https://github.com/allure-framework/allure3/pull/669), and the comments above confirm that the original problem was addressed. I’m closing this issue. If the raw rendering still occurs on a current Allure 3 version, please open a new issue with a minimal attachment and version details.

Issue #692, opened 2026-06-08
HTML attachment rendering issues – garbled Chinese text, fullscreen shows source code, and preview button does not expand properly


hzs1212:
Version 3.9.0 still has several HTML attachment rendering issues:

1. **Rendering problem**: Allure reports can render HTML attachments, but when displaying HTML format, the Chinese text background colors do not show up, and Chinese characters appear as garbled/mojibake.

<img width="1313" height="268" alt="Image" src="https://github.com/user-attachments/assets/52175d6f-bd0a-4bff-b678-a10a27d45fb4" />

2. **Fullscreen display issue**: When clicking the expand button to view in fullscreen mode, the attachment displays the HTML source code by default instead of rendering the HTML format.

<img width="1802" height="835" alt="Image" src="https://github.com/user-attachments/assets/db99dfad-c62e-4c25-89b3-39ce969f8b61" />

3. **Preview button behavior**: When in fullscreen attachment viewing mode, clicking the preview button shows the same effect as non-fullscreen mode, and only displays in a small window instead of filling the fullscreen window.

<img width="1816" height="841" alt="Image" src="https://github.com/user-attachments/assets/e02526c0-dafd-48de-9958-35bb1e83638f" />

HTML Attachment as follows:
[db1163a11e2b16f73925e7c3ae20add3.html](https://github.com/user-attachments/files/28585913/db1163a11e2b16f73925e7c3ae20add3.html)

kieran-ryan:
> Preview button behavior: When in fullscreen attachment viewing mode, clicking the preview button shows the same effect as non-fullscreen mode, and only displays in a small window instead of filling the fullscreen window.

Related to #679.

yvinokurovprogressive:
HTML rendering issue persists in v3.13. Only third of the modal renders the attachment

<img width="927" height="459" alt="Image" src="https://github.com/user-attachments/assets/ba9cf128-19f3-4003-800b-6297c1a6c3a5" />

PR #697, opened 2026-06-09
fix(web-components): fix markdown truncation and HTML attachment modal issues

### LINKED ISSUE #679: Markdown render truncated
**Describe the bug**

Allure v3.9.0 introduces markdown previews (https://github.com/allure-framework/allure3/issues/655) which improves on displaying raw source code - which in particular resolves the issue in reading tables.

The markdown preview appears to render correctly and in with suitable styling, though is obfuscated after a particular height - which means that the content is not fully visible. This also occurs when opened in full screen view.

**Expected behavior**

Full markdown contents are visible when expanded.

**Screenshots**

Image

Image

 - OS: macOS
 - Browser: Safari
 - Version: allure v3.9.0

 - OS: Windows
 - Browser: Edge
 - Version: allure v3.9.0

**Additional context**

Functionality released within v3.9.0 through https://github.com/allure-framework/allure3/issues/655.

todti:
## Summary

Fixes two bugs related to attachment preview rendering:

**#679 — Markdown render truncated**
- `MarkdownPreview` was reusing `.html-attachment-preview` CSS which applies `flex: 1; min-height: 0` to the iframe — this caused the CSS flex layout to override the dynamically-measured `height` attribute, capping the iframe at the flex container's height instead of the full content height
- Added a dedicated `.markdown-attachment-preview` class (no flex constraints) so the measured height is respected
- Changed `height="100%"` fallback (before content loads) to no attribute — browser default (~150 px) is more sensible in a block container than `100%` of an undefined parent
- Changed `overflow: hidden` → `overflow-y: auto` on `.modal-data-component` so tall markdown content scrolls within the modal

**#692 — HTML attachment rendering issues in fullscreen**
- Wired `.attachmentViewStack` / `.attachmentViewPane` into the modal's flex height chain so `HtmlPreview`'s `height: 100%` iframe fills the fullscreen modal instead of collapsing to the browser default ~150 px
- `TrAttachment`: passed current `showPreview` state when opening the modal so HTML/markdown attachments open in the same mode (preview vs source code) the user was already viewing, instead of always defaulting to source code
- `HtmlPreview`: changed blob MIME type from `text/html` to `text/html; charset=utf-8` — since `response.text()` always decodes as UTF-8 and DOMPurify strips all `<meta charset>` tags, the blob bytes are already UTF-8; the explicit charset prevents the iframe from guessing a wrong encoding if any charset hint survives sanitization

> Note: fully fixing garbled Chinese text for GBK-encoded files would require refactoring `fetchAttachment` to detect and decode using the file's actual charset rather than defaulting to UTF-8 — that's a separate issue.

Closes #679
Related to #692


kieran-ryan:
Delighted to see this - thank you for implementing! 🙌 Had been having a quick look on the same myself to understand the desired outcome (https://github.com/kieran-ryan/allure3/tree/fix/preview-render-height) and popping some additional aspects that should complete the preview functionality - having built from your branch - hope helpful!

Expand should open in preview mode by default

Screenshot 2026-06-09 at 15 19 00

Should support markdown syntax highlighting - looks like may have been a regression within a release

Screenshot 2026-06-09 at 15 37 00

Markdown preview is scrollable beyond the end of the content of the document in expand mode

Screenshot 2026-06-09 at 15 19 43

Vertical padding missing on inline preview

Screenshot 2026-06-09 at 15 18 31

kieran-ryan:
Retested and adding below observations. Using macOS with Safari in case relevant - though might retry again to ensure local build is clean.

Now observe the vertical padding on the inline preview - nice!

Screenshot 2026-06-09 at 18 23 59

Syntax highlighting is visible on the code blocks with this theme, though remains absent from the headers - which was available in earlier releases. Perhaps is a stylistic or thematic choice? Mentioning in case.

Screenshot 2026-06-09 at 18 22 01

Expanded preview mode continues to open as raw source code by default; and the scroll height when changing to preview mode continues to extend and allow scrolling beyond the content.

One additional enhancement (which perhaps could be a separate issue if requires some work) would to be provide syntax highlighting on code blocks in preview mode.

image

kieran-ryan (approved):
Awesome! Can confirm working well from local build.

- Vertical padding on inline preview
- Expand opens in preview mode by default
- Syntax highlighting on markdown
- Scroll in expand mode matches content height

Syntax highlighting of code blocks within rendered markdown desirable, though can be a standalone issue if to be addressed separately.

Also closes #431.

Looking forward to the fix release - thank you! 🚀

formaceft-93 (approved):

PR #719, opened 2026-06-18
Add generic performance metrics support to Allure reports

formaceft-93:
The main use case of the feature is performance measurements, but the format is intentionally not tied to Allure’s own benchmark phases, so other projects can publish their own numeric metrics through perf.json.
The report now reads explicit metric samples, adapts the current allure-perf-metrics.json format, and can also flatten arbitrary numeric JSON values into metric keys. Metrics are stored separately from test results, exported to widgets/metrics.json, attached as raw input for audit/debugging, and included in report history as stable key -> value pairs for trend views.
On the UI side, Awesome gets a dedicated Metrics section with grouped summary rows, current values, deltas, trends, and a selected history metric. Dashboard gets a compact metrics widget. The internal perf writer also emits an allure.total summary metric so the default history view can show an end-to-end report generation number instead of a project-specific phase.

Example perf.json:

```
{
  "version": 1,
  "name": "Build performance",
  "display": {
    "historyMetricKey": "build.total.avgMs"
  },
  "metrics": [
    {
      "key": "build.total.avgMs",
      "name": "Build total",
      "group": "Build",
      "value": 12450,
      "unit": "ms",
      "better": "lower"
    }
  ]
}
```


packages/web-awesome/src/stores/sections.ts
```diff
@@ -2,16 +2,22 @@ import { getReportOptions } from "@allurereport/web-commons";
 import { computed, effect } from "@preact/signals";
 
 import type { AwesomeReportOptions } from "../../types.js";
-import { navigateToRoot, navigateToSection, sectionRoute } from "./router";
+import { navigateToRoot, navigateToSection, SECTION_ROUTE_NAMES, sectionRoute, type SectionRouteName } from "./router";
 
 const DEFAULT_SECTION = "default";
 
-type Section = "timeline" | "charts" | "default";
+type Section = SectionRouteName | "default";
```
epszaw:
```suggestion
type Section = SectionRouteName | DEFAULT_SECTION;
```

packages/web-awesome/src/stores/reportRootTabs.ts
epszaw:
```suggestion
  const tabs: ReportRootTabId[] = [
    REPORT_ROOT_TAB.Results, 
    REPORT_ROOT_TAB.QualityGate,
    REPORT_ROOT_TAB.GlobalAttachments, 
    REPORT_ROOT_TAB.GlobalErrors,
  ];
```

packages/web-awesome/src/stores/reportRootTabs.ts
```diff
@@ -36,7 +34,9 @@ export const getAvailableReportRootTabs = (): ReportRootTabId[] => {
     tabs.push(REPORT_ROOT_TAB.Categories);
   }
 
-  tabs.push(REPORT_ROOT_TAB.QualityGate, REPORT_ROOT_TAB.GlobalAttachments, REPORT_ROOT_TAB.GlobalErrors);
+  tabs.push(REPORT_ROOT_TAB.QualityGate);
```
epszaw:
```suggestion
```

packages/web-awesome/src/stores/reportRootTabs.ts
```diff
@@ -36,7 +34,9 @@ export const getAvailableReportRootTabs = (): ReportRootTabId[] => {
...
+
```
epszaw:
```suggestion
```

packages/web-awesome/src/stores/reportRootTabs.ts
```diff
@@ -36,7 +34,9 @@ export const getAvailableReportRootTabs = (): ReportRootTabId[] => {
...
+  tabs.push(REPORT_ROOT_TAB.GlobalAttachments, REPORT_ROOT_TAB.GlobalErrors);
```
epszaw:
```suggestion
```

packages/web-awesome/src/stores/reportRootTabs.ts
```diff
@@ -36,7 +34,9 @@ export const getAvailableReportRootTabs = (): ReportRootTabId[] => {
     tabs.push(REPORT_ROOT_TAB.Categories);
```
epszaw: If you need to keep the order, use `splice` instead of `push`, like:

```ts
tabs.splice(1, 0, REPORT_ROOT_TAB.Categories);
```

packages/reader/src/perf/index.ts
```diff
@@ -0,0 +1,197 @@
+import type { MetricBetter, MetricSample } from "@allurereport/core-api";
+import type { RawGlobals, ResultsReader } from "@allurereport/reader-api";
+
+const readerId = "perf";
+const legacyFiles = new Set(["perf.json", "allure-perf-metrics.json"]);
```
epszaw: I think we can omit legacy file support because the feature couldn't be widely adopted across the users, except us. We can survive this, but we can support the actual files format only

epszaw: Due to the legacy file handling, the reader's logic looks pretty complicated.

packages/reader-api/src/reader.ts
```diff
@@ -11,6 +11,7 @@ export interface ReaderContext {
 export interface ResultsVisitor {
   visitTestResult(result: RawTestResult, context: ReaderContext): Promise<void>;
   visitCheckResult(result: AllureCheckResult, context: ReaderContext): Promise<void>;
+  visitMetrics(metrics: MetricSample[], context: ReaderContext): Promise<void>;
```
epszaw: The raw representation of the performance metrics result is `AllurePerformanceResult`. `MetricSample` is the entity we receive after the raw data enrichment

packages/sandbox/allure-results/00000000-0000-4000-8000-000000000001-perf.json
```diff
@@ -0,0 +1,26 @@
+{
```
epszaw: I propose to use a single file per result, like we do with test results

packages/plugin-api/src/qualityGate.ts
```diff
@@ -32,13 +38,17 @@ export interface QualityGateRuleState<T> {
...
-  message: (payload: { expected: T; actual: T }) => string;
+  message: (payload: { expected: T; actual: any; performance?: PerformanceConfig }) => string;
   validate: (payload: {
     expected: T;
     trs: TestResult[];
     knownIssues: KnownTestFailure[];
     state: QualityGateRuleState<K>;
+    metrics?: MetricSample[];
+    history?: HistoryDataPoint[];
+    performance?: PerformanceConfig;
```
epszaw: Why do we need performance config there?
The performance result already contains `group` and `value` to compare with the rule condition, so most likely we don't need to have access to the performance config there.

epszaw (changes requested):

packages/sandbox/allure-results/00000000-0000-4000-8000-000000000101-perf.json
```diff
@@ -0,0 +1,7 @@
+{
```
epszaw: Don't see the related changes in the `.gitignore` file, but the `allure-results` directory is ignored. Let's allow exactly these files using the `!` expression in the `.gitignore` file

packages/reader/test/resources/perf/single-performance-result.json
```diff
@@ -0,0 +1,7 @@
+{
```
epszaw: I think it's better to use a single notation – the array one. Let's remove single object notation samples and support of the notation and make array-like notation as a standard for performance result files content.

packages/web-awesome/src/stores/sections.ts
```diff
@@ -2,16 +2,22 @@ import { getReportOptions } from "@allurereport/web-commons";
...
+import { navigateToRoot, navigateToSection, SECTION_ROUTE_NAMES, sectionRoute, type SectionRouteName } from "./router";
 
 const DEFAULT_SECTION = "default";
 
-type Section = "timeline" | "charts" | "default";
+type Section = SectionRouteName | DEFAULT_SECTION;
 
 const reportOptions = getReportOptions<AwesomeReportOptions>();
-const defaultSectionFromReportOptions: Section = (reportOptions?.defaultSection as Section) ?? "default";
+const isKnownSection = (value: unknown): value is SectionRouteName =>
```
epszaw, line 12: We can a bit simplify the helper and make some kind of normalizer/adapter:

```ts
normalizeSection("known_section") // returns "known_section"
normalizeSection("unknown_section") // returns "default" section
```

In that way, we don't need to add additional checks; just pass the existing value and receive the expected one.

packages/reader/src/perf/index.ts
```diff
@@ -0,0 +1,85 @@
+import type { AllurePerformanceResult } from "@allurereport/core-api";
```
epszaw: I think, we should handle `*-perf.json` files in the `allure2` reader like we do for the other new files produced by the integrations (such as `*-globals.json`).

packages/plugin-api/src/qualityGate.ts
```diff
@@ -32,13 +32,16 @@ export interface QualityGateRuleState<T> {
 
 export type QualityGateRule<T = any, K = T> = {
   rule: string;
-  message: (payload: { expected: T; actual: T }) => string;
+  message: (payload: { expected: T; actual: any; metrics?: MetricSample[] }) => string;
```
epszaw: Why do we need to pass the metrics to the rule's message formatter? It's enough to return `epxected` and `actual` values from the `validate` function and use them in the message

packages/plugin-api/src/qualityGate.ts
```diff
@@ -32,13 +32,16 @@ export interface QualityGateRuleState<T> {
...
   validate: (payload: {
     expected: T;
     trs: TestResult[];
     knownIssues: KnownTestFailure[];
     state: QualityGateRuleState<K>;
+    metrics?: MetricSample[];
+    history?: HistoryDataPoint[];
     environment?: string;
+    currentReportUuid?: string;
```
epszaw: How do you plan to use the parameter?

packages/plugin-api/src/qualityGate.ts
```diff
@@ -32,13 +32,16 @@ export interface QualityGateRuleState<T> {
...
   rule: string;
-  message: (payload: { expected: T; actual: T }) => string;
+  message: (payload: { expected: T; actual: any; metrics?: MetricSample[] }) => string;
   validate: (payload: {
     expected: T;
     trs: TestResult[];
     knownIssues: KnownTestFailure[];
     state: QualityGateRuleState<K>;
+    metrics?: MetricSample[];
+    history?: HistoryDataPoint[];
```
epszaw: What's the purpose of the history there?

epszaw (changes requested):

packages/plugin-api/src/qualityGate.ts
formaceft-93: I renamed `history` to `previousHistory` to make the contract more explicit.

It is used only by metric delta quality gate rules, such as `metricMaxDelta` and `metricMaxDeltaPercent`. These rules compare the current metric value from `metrics` with the latest previous value for the same key from report history.

packages/reader/src/allure2/index.ts
```diff
@@ -97,6 +99,25 @@ export const allure2: ResultsReader = {
...
 
+    if (originalFileName.endsWith("-perf.json") || originalFileName === "performance.json") {
+      try {
+        const results = performanceResults(await data.asJson<unknown>());
+
+        if (results.length === 0) {
+          return false;
+        }
+
+        await visitor.visitAttachmentFile(data, { readerId });
```
epszaw: Why do we need to visit the file as attachment? We already have `results`, so we can just call `visitMetrics` with the results

packages/reader/src/allure2/index.ts
```diff
@@ -97,6 +99,25 @@ export const allure2: ResultsReader = {
...
+        await visitor.visitGlobals(rawAttachmentGlobals(originalFileName, data.getContentType()), { readerId });
```
epszaw: Same here. We're mixing the responsibility of the logic branch, but we have to process only performance metrics results there

packages/reader/src/allure2/index.ts
```diff
@@ -265,6 +286,49 @@ const processCheckResult = async (
   );
 };
 
+const finiteNumber = (value: unknown): number | undefined => (Number.isFinite(value) ? Number(value) : undefined);
+
+const normalizePerformanceResult = (result: Record<string, unknown>): AllurePerformanceResult | undefined => {
```
epszaw: If the data produced by our integrations is invalid – we need to fix the integration instead of creating similar normalizers, so I don't know, do we need these functions at all.

packages/plugin-api/src/qualityGate.ts
```diff
@@ -32,12 +32,14 @@ export interface QualityGateRuleState<T> {
 
 export type QualityGateRule<T = any, K = T> = {
   rule: string;
-  message: (payload: { expected: T; actual: T }) => string;
+  message: (payload: { expected: T; actual: any }) => string;
```
epszaw: `actual` should have the same type as `expected`, the current signature allows implementing inconsistent behavior

packages/core/src/report.ts
```diff
@@ -1081,6 +1138,13 @@ export class AllureReport {
         });
       });
       this.#finishGeneratePerfSpan();
+
+      if (await this.#ingestSelfPerfMetrics()) {
```
epszaw: I'm not sure that we need to support dynamically changing configuration exclusively for the performance feature. Let's not do that and use the configuration loaded once the report starts

epszaw (changes requested):

packages/core/src/report.ts
formaceft-93: Yes, agreed and fixed. We no longer support dynamically changing performance configuration here.

The performance config is loaded once when the report starts and passed into `DefaultAllureStore`. The `if (await #ingestSelfPerfMetrics())` branch only checks whether self-performance results were collected and added to the store. If they were, UI plugins refresh their generated files so `widgets/metrics.json` and `index.html` include the metrics before publish. It doesn’t reload or mutate configuration.

epszaw (approved):

PR #721, opened 2026-06-19
chore(plugin-testops): support `enable` property from config

formaceft-93:
The config resolver now passes the resolved plugin enabled state into plugin constructors, so enabled defaults are handled consistently.

### DIFF
--- a/packages/plugin-testops/src/plugin.ts
+++ b/packages/plugin-testops/src/plugin.ts
@@ -3,7 +3,13 @@ import { env } from "node:process";
 import { detect, isLocalCiDescriptor } from "@allurereport/ci";
 import type { CategoryDefinition, EnvironmentIdentity, TestStatus } from "@allurereport/core-api";
 import { getWorstStatus } from "@allurereport/core-api";
-import { type AllureStore, type Plugin, type PluginContext, createPluginSummary } from "@allurereport/plugin-api";
+import {
+  type AllureStore,
+  type Plugin,
+  type PluginConstructorContext,
+  type PluginContext,
+  createPluginSummary,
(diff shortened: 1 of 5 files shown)


packages/plugin-testops/src/plugin.ts
```diff
@@ -31,9 +37,21 @@ export class TestOpsPlugin implements Plugin {
...
+  constructor(
+    readonly options: TestOpsPluginOptions,
+    context: PluginConstructorContext = {},
+  ) {
+    this.#enabledByConfig = context.enabled === true;
+    this.#disabledByConfig = context.enabled === false;
+
+    if (this.#disabledByConfig) {
+      return;
+    }
```
vdvukhzhilov, line 53: looks like `this.#disabledByConfig` is only needed here, so maybe 
```suggestion
  #enabledByConfig: boolean = false;

  constructor(
    readonly options: TestOpsPluginOptions,
    context: PluginConstructorContext = {},
  ) {
    this.#enabledByConfig = context.enabled === true;

    if (context.enabled === false) {
      return;
    }
```

vdvukhzhilov (approved):

packages/plugin-testops/src/plugin.ts
formaceft-93: i got rid of this.#disabledByConfig and now it's even clearer (as for me)

vdvukhzhilov (changes requested):
looks like if `enabled` is not passed in config, the plugin will run locally anyways. And this is not desired.

vdvukhzhilov (approved):

PR #737, opened 2026-06-24
fix: only fetch commit hashes in git

vdvukhzhilov:


packages/git/src/collectGitFacts.ts
```diff
@@ -49,6 +49,23 @@ const resolveBranch = (cwd?: string): string | undefined => {
   return branchRaw;
 };
 
+const DEEPEN_FETCH_FILTERS = ["--filter=tree:0", "--filter=blob:none"] as const;
+
+const deepenFetch = (deepenBy: number, cwd?: string): void => {
+  const deepen = String(deepenBy);
+  const strategies: string[][] = [
+    ["fetch", "--deepen", deepen, DEEPEN_FETCH_FILTERS[0], "origin"],
```
delatrie: Let's also fetch from the default remote, which might differ from `origin`:

```suggestion
    ["fetch", "--deepen", deepen, DEEPEN_FETCH_FILTERS[0]],
```

delatrie (changes requested):

packages/git/src/collectGitFacts.ts
vdvukhzhilov: easy

delatrie (approved):

PR #740, opened 2026-06-26
Hide and mask sensitive parameters in report output

formaceft-93:
Allure 3 now respects hidden and masked test parameters when generating Awesome, Classic, and Allure 2 style reports. Hidden parameters are removed from report payloads, while masked parameters keep their names but show "<masked>" instead of the original value.

### DIFF
--- a/packages/core-api/src/utils/parameter.ts
+++ b/packages/core-api/src/utils/parameter.ts
@@ -1,0 +1,18 @@
+import type { TestParameter } from "../metadata.js";
+
+const maskedParameterPlaceholder = "<masked>";
+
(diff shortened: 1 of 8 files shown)


packages/core-api/src/utils/parameter.ts
```diff
@@ -0,0 +1,18 @@
+import type { TestParameter } from "../metadata.js";
+
+const maskedParameterPlaceholder = "<masked>";
+
+export const redactParameters = (parameters: readonly TestParameter[] | undefined): TestParameter[] => {
+  return (parameters ?? []).flatMap((parameter) => {
```
epszaw: If the function accepts just a list of parameters, there's no need to use `flatMap`, but just `map`

epszaw (approved):

PR #774, opened 2026-07-11
Fix stale pie chart and categories when switching environments (fixes #647)

### LINKED ISSUE #647: styles_chart in report does not auto-refresh when switching environments
**Describe the bug**
The styles_chart in the report view does not update automatically when switching between environments using the environment selector at the top. The chart continues to display stale data from the previously selected environment.

**To Reproduce**
1. Navigate to the report view
2. Observe the styles_chart displaying data for the current environment
3. Click on a different environment in the environment selector at the top
4. See that the styles_chart still shows the old data and does not refresh

**Expected behavior**
The styles_chart should automatically re-fetch and display the data corresponding to the newly selected environment — without requiring a manual page refresh.

**Screenshots**

Image

Image

After reload the page manuell we get the correct chart:
Image

**Additional context**
A manual page refresh after each environment switch resolves the display issue temporarily. This suggests the chart component is not reactively listening to environment state changes.

d-braun:
### Context

Fixes #647 

The success-rate pie chart in the report header did not update when switching
environments via the environment picker (#647). The categories tab had the
same problem, including when switching back to "All environments".

Both stores were only fetched once on app start with the initially selected
environment. The `useSignalEffect` in `packages/web-awesome/src/index.tsx`
that reacts to environment changes only re-fetched trees and stats
(`fetchEnvTreesData`, `fetchEnvStats`), so `pieChartStore`
(`widgets/<env>/statistic.json`) and `categoriesStore`
(`widgets/<env>/categories.json`) kept showing data from the previous
environment until a full page reload.

The effect now also re-fetches the pie chart and categories data on every
environment change. It additionally runs when switching back to
"All environments" (empty env id), which previously aborted early.

Other environment-dependent views (overview charts, trees, stats, quality
gate, globals, metadata) were checked and already handle environment
switching correctly — they either keep per-environment data and select it
reactively via signals, or re-fetch on change themselves.

Verified manually with the sandbox demo data (environments `foo`/`bar`):
the header pie chart and categories now update immediately on switch.
All 184 existing tests in `@allurereport/web-awesome` pass.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/web-awesome/src/index.tsx
+++ b/packages/web-awesome/src/index.tsx
@@ -90,7 +90,14 @@ const App = () => {
   useSignalEffect(() => {
     const envId = currentEnvironment.value;
 
-    if (!prefetched || !envId) {
+    if (!prefetched) {
+      return;
+    }
+
+    fetchPieChartData(envId);
+    fetchCategoriesData(envId);
+
+    if (!envId) {
       return;
     }
 
--- a/packages/web-awesome/test/index.test.tsx
+++ b/packages/web-awesome/test/index.test.tsx
@@ -1,0 +1,166 @@
+import { act, screen, waitFor } from "@testing-library/preact";
+import { epic, feature, label, story } from "allure-js-commons";
+import { beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+
+beforeEach(async () => {
+  await epic("coverage");
+  await feature("ui-components");
+  await story("App");
+  await label("coverage", "ui-components");
+});
+
+const {
+  ensureReportDataReadyMock,
+  fetchReportStatsMock,
+  fetchEnvStatsMock,
+  getLocaleMock,
+  fetchPieChartDataMock,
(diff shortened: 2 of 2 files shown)


todti:
Hi @d-braun! 
Thanks for the fix, this looks correct.I see the "Provide unit tests" checkbox is still unchecked, any chance you could add a test covering this regression (switching to an environment and back to "All environments")? 

Would be good to have since it's an easy one to accidentally break again.

d-braun:
Hey @todti, good call!
Added tests for both directions (env selected / back to "All environments") in `packages/web-awesome/test/index.test.tsx` and proved they catch the regression by running them against the pre-fix code (red without the fix, green with it).

todti (approved):

d-braun:
Hey @todti ,
just checking in on this PR — is there anything I still need to do on my side, or will the PR be merged by a maintainer? Thanks!

PR #777, opened 2026-07-15
fix pw trace viewer

### LINKED ISSUE #624: unable to open playwright trace file natively in allure3 report

formaceft-93:
fixes https://github.com/allure-framework/allure3/issues/624

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/e2e/test/allure-awesome/features/attachments.test.ts
+++ b/packages/e2e/test/allure-awesome/features/attachments.test.ts
@@ -705,6 +705,7 @@ test.describe("attachments", () => {
 
       if (popup) {
         await popup.waitForURL("https://trace.playwright.dev/", { timeout: 5_000 });
+        await expect(popup.getByText("start.test.ts")).toBeVisible({ timeout: 10_000 });
         return;
       }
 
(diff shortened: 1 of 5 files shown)


epszaw (approved):
It works, but the 2s delay can be confusing because the first thing I see is an empty trace viewer, so I can decide that the viewer doesn't work:

https://github.com/user-attachments/assets/5381a2f6-bf33-439f-ae31-40541e1d6f26

epszaw (approved):

PR #793, opened 2026-07-16
fix(web-components): preserve <head> styles when rendering HTML attachments

### LINKED ISSUE #692: HTML attachment rendering issues – garbled Chinese text, fullscreen shows source code, and preview button does not expand properly

todti:
## Summary
- `HtmlPreview` sanitized HTML attachments with DOMPurify's default (`WHOLE_DOCUMENT: false`) config, which drops the entire `<head>` — including `<style>` blocks, `<link rel="stylesheet">`, and `<meta charset>`. Styled HTML attachments (e.g. text with custom background/foreground colors) rendered as unstyled plain text.
- Adds `sanitizeHtmlDocument()` in `web-commons`, which keeps `<head>` content while still forbidding `<base>` and the `http-equiv` attribute — both allow document-hijacking/redirect behavior without needing script execution, so they're stripped even though the preview iframe already has no `allow-scripts`.
- `HtmlPreview` now uses `sanitizeHtmlDocument()` instead of the unconfigured `sanitize()`.

Fixes #692

### DIFF
--- a/packages/web-commons/src/sanitize.ts
+++ b/packages/web-commons/src/sanitize.ts
@@ -32,4 +32,10 @@ const HTML_DOCUMENT_SANITIZE_CONFIG = {
  * Sanitize a full HTML document (doctype/head/body) for rendering in a preview iframe,
  * preserving `<head>` content such as `<style>` and `<meta charset>`.
  */
-export const sanitizeHtmlDocument = (html: string) => sanitize(html, HTML_DOCUMENT_SANITIZE_CONFIG);
+export const sanitizeHtmlDocument = (html: string) => {
+  const sanitized = sanitize(html, HTML_DOCUMENT_SANITIZE_CONFIG);
+  // DOMPurify has no concept of a doctype node, so WHOLE_DOCUMENT always drops it. Re-add it
+  // unconditionally: without it the iframe renders in quirks mode, which affects box sizing and
+  // other layout rules regardless of what the original attachment declared.
+  return sanitized ? `<!DOCTYPE html>${sanitized}` : sanitized;
+};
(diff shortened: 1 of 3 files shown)


packages/web-commons/src/sanitize.ts
```diff
@@ -13,3 +13,23 @@ const IFRAME_SANITIZE_CONFIG = {
...
+  // attachment, so both are dropped even though the sandboxed iframe has no allow-scripts.
+  FORBID_TAGS: ["base"],
+  FORBID_ATTR: ["http-equiv"],
+};
+
+/**
+ * Sanitize a full HTML document (doctype/head/body) for rendering in a preview iframe,
+ * preserving `<head>` content such as `<style>` and `<meta charset>`.
+ */
+export const sanitizeHtmlDocument = (html: string) => sanitize(html, HTML_DOCUMENT_SANITIZE_CONFIG);
```
formaceft-93: this helper will cut "<!DOCTYPE html>" line, so [here](https://github.com/allure-framework/allure3/pull/793/changes#diff-1292721e2cb06cef2a1926e5ccfab19cf8f22f987ded1a7dbef9625700b9ef48R6) we will lose it, but seems its okay, yea?

formaceft-93 (approved):

formaceft-93 (approved):

epszaw (approved):

PR #801, opened 2026-07-17
fix/html-attachment-preview-scroll

anilreddy:
fix(web-components): make HTML attachment preview scrollable instead of clipping content

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


todti:
Thanks for the fix! I tried to reproduce the clipping issue against a real build (sandbox fixtures + `allure generate`, not a synthetic mockup) to make sure the change actually addresses it, and found something worth flagging.

Inspecting the live DOM, `.html-attachment-preview` never actually grows past `150px` (the iframe's UA-default replaced-element height), both before and after this patch:

| element | height |
|---|---|
| `.html-attachment-preview` | 150px |
| `.attachmentViewPane` | 150px |
| `.attachmentViewStack` | 150px |
| modal content area above them | 600–1000px+ (mostly unused space) |

`.html-attachment-preview` declares `height: 100%`, but its ancestors `.attachmentViewPane` and `.attachmentViewStack` are `display: block` with no explicit height — they just shrink-wrap to their content. So `100%` has nothing to resolve against, collapses, and the iframe falls back to its default 150px box. That's the real source of the clipping.

I live-patched the DOM with exactly this PR's CSS (`overflow-y: auto` on the container + `min-height: 100%` on the iframe) and the wrapper stayed at 150px — `min-height: 100%` can't grow relative to a parent that has no resolved height of its own, so long attachments are still clipped after the change.

For comparison, HTML attachments that don't hide their own overflow already scroll fine natively inside the iframe today (`iframe.contentWindow.scrollBy` works), so this only affects content that's taller than the 150px fallback box regardless of the attachment's own CSS.

It might be worth fixing the height further up the chain (e.g. giving `.attachmentViewPane`/`.attachmentViewStack` a resolvable height) rather than (or in addition to) the two-line change here — otherwise the preview will still be stuck at 150px.

anilreddy:
You're completely right, thank you for catching this — I re-tested against a real `allure generate` build (not a mockup) and reproduced your exact numbers: `.attachmentViewPane`/`.attachmentViewStack` are `display: block`
with no resolved height, so `height: 100%` on `.html-attachment-preview` had nothing to resolve against. My original two-line patch didn't touch that chain at all, so it correctly did nothing.

I dug further and found the actual root cause is two separate problems depending on where the attachment renders:

1. **Fullscreen modal** — exactly as you found. `.modal-data-component` *does* have a real resolved height (measured live: 581.594px), but it never reaches the iframe because `.attachmentViewPane`/`.attachmentViewStack`
break the flex chain with `display: block` and no height participation.

2. **Inline (in-page) preview** — there's no ancestor viewport height to inherit at all here (parents just shrink-wrap to content, by design). This isn't a "give it 100%" problem — it needs the same auto-resize technique `MarkdownPreview.tsx` already implements (`getIframeContentHeight` + `onLoad`). I found the reason `HtmlPreview` can't just borrow that as-is: `.html-attachment-preview iframe` has `flex: 1 1 0%` in the CSS, and in a flex column, `flex-basis` wins over an explicit `height` — so even setting `iframe.style.height` via the same onLoad pattern silently does nothing until that `flex` rule is also removed. `MarkdownPreview`'s container avoids this because the *flex item* there is `.markdown-attachment-preview` itself, not the iframe — the iframe inside it is a plain block box that respects an explicit height.

I verified both fixes together, live, against a real generated report (Allure 3.14.3, same as your repro): inline preview now auto-grows to full content height (958px in my 40-row test fixture, zero scroll needed), and the fullscreen modal correctly grows to its real available space (582px) with the remainder reachable by native iframe scroll.

Updated PR below replaces my original one-liner.

anilreddy:
@todti I have added unit testcases as well please check once

todti:
Heads up, this doesn't build right now — both `HtmlPreview.tsx` and `styles.scss` have syntax errors in the latest commit (looks like some leftover/unfinished edits). Mind taking another look?

anilreddy:
@todti Sorry about that — my mistake, I was hand-writing diff hunks and one of them had a context-line mismatch that corrupted the file. I re-pulled the exact current `main` versions of both files, applied the changes with a scripted exact-string replace (so there's no manual-retyping risk this time), and verified each file compiles before posting:

- `Attachment/styles.scss` → compiled with `sass`, diff against the
  original compiled CSS shows exactly the 3 intended rule changes, nothing
  else.
- `Modal/styles.scss` → compiled with `sass` cleanly. Also noticed while
  fixing this: the existing `:global(.attachmentViewStack)` /
  `:global(.attachmentViewPane)` overrides in this file are dead code —
  `:global()` targets a literal unscoped class name, but `Attachment.tsx`
  applies these via CSS Modules (`styles.attachmentViewStack`), which
  compiles to a hashed class like `styles_attachmentViewStack__QMID9`,
  never the bare name. They've never matched anything. Removed them since
  the `Attachment/styles.scss` fix supersedes what they were trying to do,
  and added one working override instead (`.html-attachment-preview
  iframe { height: 100% !important }`) so the modal correctly forces the
  iframe back to filling its container instead of the onLoad-measured
  content height.
- `HtmlPreview.tsx` → type-checked with `tsc --strict --noEmit` (zero
  errors) and bundled with `esbuild` (zero errors) in addition to the 8
  Vitest unit tests from my earlier comment, which still all pass against
  this exact file.

Updated diffs below replace everything from my earlier comments.

anilreddy:
@todti can you please let me know how you are following the formatting. I will fix the test of mine.

Checking formatting...

packages/web-components/src/components/Attachment/HtmlPreview.test.tsx (0ms)

Format issues found in above 1 files. Run without `--check` to fix.
Finished in 1591ms on 1305 files using 4 threads.

anilreddy:
@todti am not sure this is the fix but I did the following:

What I verified before handing this back
* Downloaded the repo’s exact .oxfmtrc.jsonc (printWidth: 120, singleQuote: false, trailingComma: "all", arrowParens: "always", experimentalSortImports).
* Installed the exact pinned oxfmt@0.52.0 from package.json’s devDependencies.
* Ran oxfmt --check against the file in isolation and alongside its real sibling files (HtmlPreview.tsx, HttpAttachment.test.tsx) in a reconstructed directory tree — all pass, 0 issues.
* Scanned every byte of the file for non-ASCII characters — found and eliminated the one em-dash; confirmed HtmlPreview.tsx was already clean.
* Checked line endings (LF, no CRLF), trailing whitespace, BOM, and final-newline — all clean.

anilreddy:
@todti can you please check for last time and merge this change?

anilreddy:
@todti I see the following:

Looking at the job breakdown for this run:

- Lint (yarn lint + yarn format:check) → success
- Build on ubuntu-latest (yarn test) → success
- Build on macos-latest (yarn test) → success
- Build on windows-latest (yarn test) → success
- Build report → "Allure Report Official" step → failure

The only failing step is `Allure Report Official`, which uses `allure-framework/allure-action@v0` and requires `ALLURE_SERVICE_ACCESS_TOKEN` / `ALLURE_TOKEN` / `ALLURE_PROJECT_ID` secrets plus elevated `contents: write` / `pull-requests: write` permissions. GitHub Actions doesn't expose secrets or elevated permissions to workflows triggered from a fork PR (security boundary enforced by GitHub itself), so this step can't succeed on any external contributor's PR regardless of code changes — it looks like a maintainer-only/internal step. All 4 jobs that actually build, lint, format-check, and test the code passed on all three OSes.

Let me know if there's anything else you'd like me to adjust on the actual code changes — happy to take another pass if there's a real issue I'm missing.

anilreddy:
@todti sorry for the reminder can you please check this once?

anilreddy:
@todti can you please let me know if I need to close this PR and raise fresh one if you feel it is not good for merging. Can you please reply? This is the final time run the build else will re-push the PR again by closing it

PR #804, opened 2026-07-20
csv attachments as table

formaceft-93:
image

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/web-components/package.json
```diff
@@ -65,6 +65,7 @@
     "d3-array": "^3.2.4",
     "d3-axis": "^3.0.0",
     "d3-brush": "^3.0.0",
+    "d3-dsv": "^3.0.1",
```
epszaw: We can draw the table with no dependencies.
Consider removing the package and changing the current implementation to a self-authored one (render html table)

epszaw (changes requested):

epszaw (approved):

PR #813, opened 2026-07-22
Complete known issues feature

epszaw:
#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/core/src/store/store.ts
```diff
@@ -1728,17 +1840,22 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
+        testResult.known = false;
+        testResult.quarantine = false;
+        continue;
       }
 
-      existingKnown.push(...knownFailures);
-    });
-    this.#rebuildRetrySubstore();
+      testResult.known = Boolean(testResult.known || knownHistoryId);
+      testResult.quarantine = Boolean(!testResult.known && (testResult.quarantine || quarantineHistoryId));
```
formaceft-93: Looks like the restore path still keeps fixed tests in quarantine. testResult.quarantine here is set without checking the restored result status

allurerc.mjs
```diff
@@ -43,31 +45,14 @@ const config = {
...
-  variables: {
-    env_variable: "unknown",
-  },
-  environments: {
-    foo: {
-      variables: {
-        env_variable: "foo",
-        env_specific_variable: "foo",
-      },
-      matcher: ({ labels }) => labels.some(({ name, value }) => name === "env" && value === "foo"),
```
formaceft-93, line 55: do we need this change?

packages/sandbox/quarantine.json
```diff
@@ -0,0 +1 @@
...e-vitest/dist/matchers.js:309:24)\n    at Proxy.<anonymous> (file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/__virtual__/allure-vitest-virtual-49d332e202/0/cache/allure-vitest-npm-3.9.0-fc8d243ef2-07dca0ebd5.zip/node_modules/allure-vitest/dist/matchers.js:343:16)\n    at /Users/epszaw/Code/Work/qameta/allure3/packages/sandbox/test/bulk.spec.ts:101:26\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/cache/@vitest-runner-npm-4.1.8-38f4df6de4-278d148212.zip/node_modules/@vitest/runner/dist/chunk-artifact.js:1903:20","expected":"false","actual":"true"}},{"historyId":"4649420b4ed2623d6bb713508bbcd5cf.d41d8cd98f00b204e9800998ecf8427e","error":{"message":"java.net.ConnectException: Connection refused","trace":"Error: java.net.ConnectException: Connection refused\n    at /Users/epszaw/Code/Work/qameta/allure3/packages/sandbox/test/bulk.spec.ts:97:19\n    at processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/cache/@vitest-runner-npm-4.1.8-38f4df6de4-278d148212.zip/node_modules/@vitest/runner/dist/chunk-artifact.js:1903:20"}}]
```
formaceft-93: same question - do we need this one? 
at least it has your local env 'epszaw' paths, but as for demo looks fine

packages/core/src/report.ts
```diff
@@ -1120,6 +1157,14 @@ export class AllureReport {
         }
       }
 
+      if (this.#quarantinePath) {
```
formaceft-93: because of [default](https://github.com/allure-framework/allure3/pull/813/changes#diff-4114b8f08217be27efd35a3891857cfabc6eb75aee8ab39ba304e4d1b1095323R329) path will be always set (except we manually null it?) and every report generation will touch working tree files because quarantine lives in root

formaceft-93 (changes requested):

allurerc.mjs
epszaw: Yes, the properties are pretty useless

packages/sandbox/quarantine.json
epszaw: Removed

packages/core/src/report.ts
epszaw: Removed the default values because the parameters are optional

packages/core/src/store/store.ts
epszaw: Done

formaceft-93 (approved):

quarantine.json
formaceft-93, line 1: we don't need it anymore

packages/core/src/known.ts
```diff
@@ -1,15 +1,91 @@
...
+  knownIssues: KnownIssuesConfig | undefined,
+  environmentId?: string,
+): KnownTestFailure | undefined => {
+  if (!testResult.historyId || !hasKnownIssueRules(knownIssues)) {
+    return undefined;
+  }
+
+  const rule = knownIssues!.rules.find(
+    (item) =>
+      messageMatches(item, testResult) && testCaseMatches(item, testResult) && environmentMatches(item, environmentId),
```
formaceft-93: we have "retryHash" field in KnownIssueDescriptor, however we don't use it. Should we have additional matcher here by retryHash or it's for future purpose?

packages/core/src/known.ts
```diff
@@ -24,20 +100,19 @@ export const readKnownIssues = async (knownIssuePath: string): Promise<KnownTest
...
-    }));
+  await mkdir(dirname(path), { recursive: true });
+  await writeFile(path, `${JSON.stringify(knownIssues)}\n`, "utf-8");
+};
+
+export const writeKnownIssues = async (store: AllureStore, knownIssuesPath?: string) => {
+  const knownIssues = await store.allKnownIssues();
 
-  await writeFile(knownIssuesPath, JSON.stringify(knownIssues));
+  await writeKnownIssueRecords(knownIssues, knownIssuesPath);
```
formaceft-93, line 124: "known-issues.json" will be created even when no failed result matched any rule, is it intended?

packages/core/src/known.ts
```diff
@@ -1,15 +1,91 @@
...
+    return true;
+  }
+
+  const message = testResult.error?.message;
+
+  return typeof message === "string" && new RegExp(rule.messageRegexp).test(message);
+};
+
+const testCaseMatches = (rule: KnownIssueDescriptor, testResult: TestResult): boolean => {
+  return !rule.testCaseId || rule.testCaseId === testResult.testCase?.id;
```
formaceft-93: testCaseId rules are matched against the internal hashed testCaseId (now the raw one) - will it be clear for config authors that they should use md5(testCaseId) there?

formaceft-93 (commented):
a few questions

quarantine.json
epszaw: 👍

packages/core/src/known.ts
epszaw: Sure, we should

packages/core/src/known.ts
epszaw: It's ok, when the known issues are configured

packages/core/src/known.ts
epszaw: Done

packages/core/src/known.ts
epszaw: I slightly changed the logic. I propose to allow checking not only the test case id, but also the allure id and external id too. In that way user have a wide choice and don't need to think abount any kind of hashes

formaceft-93 (approved):

PR #816, opened 2026-07-23
Improve realtime watch reliability and add scoped rerun

todti:
## Summary

- **Scoped rerun in `allure run --watch`** (new): when a spec file changes, only that file is rerun instead of the entire suite — makes watch mode practical for large e2e suites where rerunning everything on every save was too slow.
- **Reliable shutdown for `allure run --watch`**: Ctrl+C now reliably kills the full test process tree (SIGTERM, then SIGKILL if the process doesn't exit in time), instead of sometimes hanging or leaving orphaned processes behind.
- **Crash isolation for `allure run --watch`**: if a single rerun fails unexpectedly (e.g. a spawn error), the watch session now logs it and keeps watching, instead of the whole process crashing.
- **Fixed EMFILE crashes in `allure run --watch`**: source-file watching now skips `node_modules`, dotfiles/dot-directories, and build output by default, instead of exhausting the OS file-descriptor limit on large repos.
- **Dynamic result discovery in `allure watch`**: `allure-results` directories that don't exist yet at startup (e.g. a package that hasn't run its tests yet) are now picked up automatically once they appear, instead of only ever watching the directories found at launch.
- **Interruptible shutdown for `allure watch`**: Ctrl+C now interrupts an in-progress directory scan immediately instead of waiting for the whole scan to finish first.
- **New `--new-only` flag for `allure watch`**: skips the existing backlog of results on disk at startup and only reacts to results written after the watch session started.
- **Fixed runaway terminal output in watch mode's progress display**: the live `failed/broken/passed/total` line no longer duplicates or corrupts itself when re-rendered rapidly or when other plugins log output at the same time.
- **TestOps: stopped redundant uploads**: global attachments and global errors were being re-uploaded on every single realtime update cycle; now each is uploaded only once.
- **TestOps: added a heads-up log message** announcing newly found test results before they start uploading, for better visibility into what's happening during a watch session.


packages/directory-watcher/src/index.ts
```diff
@@ -3,6 +3,18 @@ import console from "node:console";
 import { watch as chokidarWatch } from "chokidar";
 import type { EventName } from "chokidar/handler.js";
 
+// large, essentially never-relevant directories that can exhaust the fs-watch handle limit (EMFILE)
+const DEFAULT_IGNORED_DIR_NAMES = ["node_modules", "dist", "build", "out", "coverage"];
```
epszaw: We can try to read the gitignore file in the cwd because chokidar supports the same format patterns

packages/directory-watcher/src/index.ts
```diff
@@ -3,6 +3,18 @@ import console from "node:console";
...
+
+// dot-directories (.git, .yarn, .cache, .codegraph, …) are ignored wholesale too — same fd-limit
+// reasoning, plus some contain things fs.watch can't handle (e.g. a unix socket)
+const isDotSegment = (segment: string): boolean => segment.length > 1 && segment.startsWith(".");
```
epszaw: As I wrote above, we can read the gitignore file contents and add the `.git` directory to the final list. I think it's not a good idea to ignore everything with the `.` prefix because users might store test data in similar directories. We should watch everything under the given path, and I guess there's a small chance to have the listed directories in the specified path

packages/cli/src/commands/watch.ts
```diff
@@ -64,14 +66,27 @@ export class WatchCommand extends Command {
     description: "The port to serve the reports on (default: random port)",
   });
 
+  newOnly = Option.Boolean("--new-only", {
```
epszaw: Let's use the mode by default in the `watch` command, but allow to pass `false` value in the CLI arguments

packages/cli/src/commands/commons/run.ts
```diff
@@ -233,6 +241,204 @@ export const runTests = async (params: {
...
+      return;
+    }
+
+    const specPath = next.value;
+    const relativeSpec = relative(cwd, specPath);
+
+    pendingSpecs.delete(specPath);
+    running = true;
+
+    try {
```
formaceft-93: This try/catch doesn't cover childProc spawn errors - terminationOf function inside only waits for the 'exit' event, so when a failed spawn emits error event and it crashes the whole watch session.

Run "yarn workspace allure start run --watch --cwd /tmp/some-fixture -- you-will-be-surprised" then edit matching spec file => we got "Unhandled 'error' event"

Should we cover this one?

packages/cli/src/commands/watch.ts
```diff
@@ -141,14 +156,62 @@ export class WatchCommand extends Command {
...
+
+          perDirectoryWatchers.delete(deletedDir);
+        }
+
+        for (const newDir of newDirectories) {
+          if (perDirectoryWatchers.has(newDir)) {
+            continue;
+          }
+
+          const watcher = newFilesInDirectoryWatcher(
```
formaceft-93, line 186: For every new directory (even created after watch command started) command 'allure watch --new-only' drops/ignores results created inside dynamically discovered result directories.

like run 'allure watch --new-only' in a cwd with no 'allure-results', then create "pkg/allure-results/result.json" and after shutdown total will be zero (however result can have any amounts of tr)

allure watch without new-only works fine for this case

packages/cli/src/commands/run.ts
```diff
@@ -148,12 +163,28 @@ export class RunCommand extends Command {
...
     const withRerun = maxRerun > 0;
-    const withQualityGate = !!config.qualityGate && !withRerun;
+
+    if (this.watch && this.once) {
+      throw new UsageError("--watch and --once cannot be used together");
+    }
+
+    const ci = detect();
+    const isCI = !isLocalCiDescriptor(ci) || env.CI === "true" || env.CI === "1";
+    const shouldWatch = this.once ? false : (this.watch ?? !isCI);
```
formaceft-93: because of this conditions "allure run -- npm test" now by default will run locally with watch (in watch mode) - looks like it unexpected?

formaceft-93 (changes requested):

packages/cli/src/commands/commons/run.ts
todti: Good catch — fixed in `terminationOf` (`packages/cli/src/utils/process.ts`), which every caller here routes through: it now also listens for the child process's `error` event and resolves instead of leaving it unhandled, so a spawn failure surfaces as a failed run instead of crashing the whole watch session. Added a regression test in `test/process.test.ts`.

packages/cli/src/commands/watch.ts
todti: Confirmed and fixed: `ignoreInitial` was applied to every dynamically discovered directory, not just the ones present at the very first scan. Now `--new-only` only skips the backlog for directories seen on the initial discovery round; directories discovered later always get fully scanned. Added a regression test in the new `test/commands/watch.test.ts`.

packages/cli/src/commands/run.ts
todti: That's intentional, not a bug — the PR makes `--watch` the default for local runs (CI is auto-detected via `@allurereport/ci`, so CI runs are unaffected), the same way tools like `vitest`/`jest --watch` behave outside CI. `--once` is provided as the explicit opt-out for anyone who wants a single local run. No change made here.

epszaw: We can't use `watch` by default because we use the command to execute tests. If we do the change, it'll break every single existing workflow using the `allure run` command, so it's better to use `watch` only when the flag is provided explicitly

epszaw (changes requested):

packages/directory-watcher/src/index.ts
todti: Done — the default ignore list now also reads plain directory/file names out of the watched directory's .gitignore (packages/directory-watcher/src/index.ts). Kept it simple (no glob/negation support) since chokidar's own ignore matching already covers most of that; can extend if richer patterns turn out to matter.

packages/directory-watcher/src/index.ts
todti: Agreed — dropped the blanket dot-prefix check. Now only node_modules/dist/build/out/coverage/.git are ignored unconditionally, plus whatever the directory's own .gitignore lists.

packages/cli/src/commands/watch.ts
todti: Done — --new-only now defaults to true, with --no-new-only to opt back into ingesting the existing backlog.

packages/cli/src/commands/run.ts
todti: Fair point, reverted — allure run --watch is explicit-flag-only again, no CI auto-detection, no --once.

formaceft-93 (approved):

formaceft-93:
allure run --watch no longer works - it is okay?

packages/cli/src/commands/watch.ts
```diff
@@ -68,14 +70,27 @@ export class WatchCommand extends Command {
     description: "Don't clear terminal output on the data refresh",
   });
 
+  newOnly = Option.Boolean("--new-only", true, {
```
formaceft-93, line 73: Is this default change intentional? `--new-only` is currently enabled by default, so plain `allure watch` now skips existing results on startup.

packages/plugin-testops/src/plugin.ts
```diff
@@ -193,10 +194,12 @@ export class TestOpsPlugin implements Plugin {
   }
 
   async #uploadGlobalErrors(store: AllureStore) {
-    const results = await store.allGlobalErrors();
+    const allResults = await store.allGlobalErrors();
+    const lastResult = allResults.at(-1);
```
formaceft-93: This does not fully dedupe global errors. Unlike global attachments, global errors are not tracked after a successful upload; the code only uploads `allGlobalErrors().at(-1)`. That means the same last error can be re-uploaded on every realtime update, and if several new global errors appear between updates, only the last one is uploaded while the earlier ones are dropped. We should track successfully uploaded global errors and upload all not-yet-uploaded ones.

packages/plugin-testops/src/plugin.ts
```diff
@@ -360,16 +370,22 @@ export class TestOpsPlugin implements Plugin {
...
       if (stage === "done") {
-        this.#logger.info("No test results to upload");
+        this.#logger.verbose("No test results to upload");
       }
 
       return;
     }
 
+    if (stage === "update") {
+      this.#logger.verbose(
```
formaceft-93, line 404: This heads-up message is currently hidden in realtime mode. `start()` sets the logger level to `info` when `context.realTime` is true, but this message is emitted with `logger.verbose()`, so users will not see it during watch uploads. The new test also fails locally with no `console.info` calls. Should this be `logger.info(...)` instead?

formaceft-93 (changes requested):

epszaw:
> allure run --watch no longer works - it is okay?

It's ok because the previous implementation was unstable, and I decided to remove it due to the instability with different test runners and frameworks

packages/plugin-testops/src/plugin.ts
epszaw: It's intended behavior because we don't need to see information about every upload in watch mode - the information just pollutes the log.
The single thing we are interested in is the link to the launch.

packages/cli/src/commands/watch.ts
epszaw: Yes, it's intentional. In my opinion, we should run the watch mode only when we're working actively on a certain set of tests, so the rest of the tests are not so interesting for us.
At least, we can try to ship the logic as an experiment and roll back the original behavior in case something goes wrong.

packages/plugin-testops/src/plugin.ts
formaceft-93: Follow-up after discussing this with @epszaw: instead of trying to dedupe global errors in realtime updates, maybe we should skip them in `update()` and upload them only from `done()`. Global errors don't currently have stable ids, and `done()` runs once, so this avoids duplicate uploads without inventing synthetic ids in TestOps.

epszaw: Done. Decided to upload globals only in the `done` stage in the real-time mode.

formaceft-93 (approved):

epszaw (approved):

PR #818, opened 2026-07-23
Retry and rate-pace TestOps uploads

todti:
## Summary

- **TestOps upload error classification** — `errors.ts` classifies upload failures into transient (5xx/429/timeout), recoverable ("launch is closed"), or terminal (auth/validation/payload/conflict) kinds, recognizing both raw HTTP errors and the `KnownError`/`UnknownError` wrappers that `@allurereport/service`'s HTTP client actually throws. `withUploadRetry()` retries only the retryable kinds with exponential backoff, and is wired into all four TestOps upload call sites (test results, global attachments, global errors, quality gate).
- **Closed-launch reopening** — a new `reopenClosedLaunch` option reopens a launch TestOps reports as closed instead of failing the upload outright.
- **Upload rate pacing** — `uploadPacer.ts` adds a leaky-bucket pacer with independent per-window budgets for requests, files, and bytes sent to TestOps, applied at every upload call site (global attachments, per-result attachments/fixtures, test results, quality gate). Enabled by default with sane defaults (20 req/s, 1000 files/s, 1 GiB/s). Can be disabled via `uploadRateLimit: false`, or tuned via an explicit `uploadRateLimit` config.
- Also fixes an `open.ts` command that only handled `SIGINT` for temp-report cleanup — it now handles `SIGTERM` too, via a small shared `utils/signals.ts` helper (`notifySignals`/`waitForAbort`) with conventional exit codes (130/143) and a graceful-shutdown deadline.

## Test plan

- [x] `yarn vitest run` in `packages/plugin-testops` — all tests pass
- [x] `tsc --noEmit` — clean
- [x] `oxlint` — no new warnings introduced

### DIFF
--- a/packages/plugin-testops/src/plugin.ts
+++ b/packages/plugin-testops/src/plugin.ts
@@ -378,10 +378,6 @@ export class TestOpsPlugin implements Plugin {
         await this.#reopenLaunchIfClosed(error);
       };
 
-      // uploadTestResults already retries each chunk in place (see client.ts) so an already
-      // acknowledged chunk is never resent. This outer retry only fires if a chunk exhausts its
-      // own retries, ponytail: rare case, still resends earlier chunks; fix with cross-call
-      // upload-id tracking if that turns out to matter in practice.
       const uploadedTrs = await withUploadRetry(
         () =>
           this.#client.uploadTestResults({
--- a/packages/plugin-testops/src/plugin.ts
+++ b/packages/plugin-testops/src/plugin.ts
@@ -15,6 +15,7 @@ import { uniqBy, stubTrue } from "lodash-es";
 import { bold } from "yoctocolors";
 
 import { TestOpsClient } from "./client.js";
+import { withUploadRetry } from "./errors.js";
 import { LaunchGitFlow, resolveGitFlowOptions } from "./gitFlow/index.js";
 import { Logger } from "./logger.js";
 import type { TestOpsPluginTestResult, TestOpsPluginOptions, UploadCategory } from "./model.js";
@@ -66,6 +67,7 @@ export class TestOpsPlugin implements Plugin {
       launchName,
       launchTags,
       autocloseLaunch = true,
+      uploadRateLimit,
     } = resolvePluginOptions(options);
 
     // don't initialize the client when some options are missing
@@ -75,6 +77,7 @@ export class TestOpsPlugin implements Plugin {
         baseUrl: endpoint,
         accessToken,
         projectId,
+        uploadRateLimit,
       });
       this.#launchName = launchName;
       this.#launchTags = launchTags;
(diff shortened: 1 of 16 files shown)


packages/plugin-testops/src/plugin.ts
```diff
@@ -321,13 +373,22 @@ export class TestOpsPlugin implements Plugin {
...
+        () =>
+          this.#client.uploadTestResults({
+            attachmentsResolver: attachmentsResolverFactory(store),
+            fixturesResolver: fixturesResolverFactory(store),
+            environments,
+            trs: trsToUpload,
+            onProgress: () => incrementProgress(),
+          }),
+        {
+          onRetry: async (error, attempt) => {
```
formaceft-93: try + retry here are wrapping full upload however if 2-3d chunk will fail for example [here](https://github.com/allure-framework/allure3/pull/818/changes#diff-e7f2163908043b20799be2dc76db154c1870a0aa16bf98d775c39ec74c8bc56cR457) - retry will duplicate them, is it okay? I mean we should be sure that testops know how to dedup results in this case

formaceft-93 (approved):

packages/plugin-testops/src/plugin.ts
todti: Fixed — retry is now scoped per chunk instead of the whole upload call. Each chunk's POST to /api/upload/test-result retries independently, so a later chunk failing no longer causes earlier, already-acknowledged chunks to be resent.

todti: Checked this — the server does not dedup by our uuid on re-upload. Test result creation resolves conflicts by internal TEST_RESULT.ID, not by the uuid we send; a repeated chunk gets a brand-new row unless it happens to match a preallocated result via job-run reconciliation (unrelated mechanism). So yes, if a chunk's request actually lands server-side but we never see the response and retry it, that would create a duplicate today. Worth tracking as a follow-up — likely either an idempotency key on the upload endpoint, or having the client only retry on errors that are safe to assume as "not applied" (e.g. connection refused before the request went out) rather than any transient failure.

formaceft-93: absolutely!

formaceft-93: > Fixed — retry is now scoped per chunk instead of the whole upload call. Each chunk's POST to /api/upload/test-result retries independently, so a later chunk failing no longer causes earlier, already-acknowledged chunks to be resent.

Do we actually need outer retry now, when we cover everything inside operations? As I see we still has another case here:

1. chunk 1 POST -> success
2. chunk 2 POST -> 503
3. inner retry chunk 2 -> 503
4. inner retry chunk 2 -> 503
5. inner retry chunk 2 -> 503
6. client.uploadTestResults throws outside
7. outer retry в plugin.ts run full client.uploadTestResults again
8. chunk 1 POST -> success second time

todti:

Issue #822, opened 2026-07-24
feat: make Status Dynamics chart bars clickable to navigate to historical reports


hasanalpzengin:
### What problem should this solve?

The Status Dynamics chart on the dashboard shows test run history across multiple executions, but the bars are purely visual — there is no way to navigate to the corresponding historical report from the chart. Users who spot a regression or an anomalous run in the trend have to manually locate the report elsewhere, breaking the investigative flow.

### What would you like to happen?

Clicking a bar in the Status Dynamics chart should navigate the user to the corresponding historical report. Specifically: 
- Each history bar should be clickable and open the report URL associated with that history data point
- The current run bar (id: "current") should either be non-clickable or navigate to the current report
- The url field already present on HistoryDataPoint in core-api should be propagated through StatusDynamicsChartData (currently it is dropped at generation time in generateStatusDynamicsChart.ts) and used as the navigation target
- The BarChart component already exposes an onBarClick prop wired to nivo's onClick — it just needs to be passed by StatusDynamicsChartWidget

### Alternatives considered

- Tooltip with a link instead of direct click navigation — less discoverable and adds an extra interaction step; direct click is more intuitive for a chart that already implies navigability through its time-series structure
- Using the existing TrendChartWidget click handling — TrendChartWidget has a handleSliceClick that sets selected slice state, but the result is unused and the widget is not rendered on the Charts pages; extending it would require more invasive changes
- Keeping bars non-clickable and adding a separate report list below the chart — already partially covered by other UI surfaces; doesn't solve the in-context navigation problem

### Additional context

Potential Solution:
1. packages/charts-api/src/types.ts — add url?: string to the StatusDynamicsChartData data item type
2. packages/web-commons/src/charts/generateStatusDynamicsChart.ts — pass point.url through when mapping HistoryDataPoint to chart data items
3. packages/web-components/src/components/Charts/StatusDynamicsChartWidget/index.tsx — pass onBarClick to BarChart, extracting the url from the clicked datum and navigating to it

The same url propagation gap exists in other history-based charts (StatusTransitionsChartData, DurationDynamicsChartData) and could be addressed consistently in a follow-up.

geoL86:
Hey, if I'm not mistaken this feature was presented on allure2 reports. I think it will be useful here too.

geoL86:
Hey @hasanalpzengin it seems your fix "URL handling for history points" was released as a part of 3.15.0 Could you guide me, how to actualy generate or add this url to report? 
[...]

PR #826, opened 2026-07-25
feat(web-components): add URL handling for history points and update (#822)

hasanalpzengin:
Add clickable history points with URL navigation in Status Dynamics Chart (#822)
    
### Context
    
    The Status Dynamics Chart previously displayed history points without any way to navigate to the report
    each point represents. This change makes the chart bars interactive so users can jump directly to a historical run.
    
    - Added an optional url field to history points across the data model: StatusDynamicsChartData in both
    charts-api and web-commons, the StatusDynamicsChartWidget props/types, and the generateStatusDynamicsChart mapper (which now carries point.url through).
    - Threaded an onBarClick handler through BarChart → BarChartItemHoverLayer →
    BarChartItemHoverArea, wiring an onClick on the hover rect and showing a pointer cursor when a click handler is present.
    - StatusDynamicsChartWidget now opens the history point's url in a new tab (_blank,
    noopener,noreferrer) on bar click, skipping the "current" point and any point without a URL.
    - Added feature test statusDynamicsChartBarClick.test.ts covering the bar-click / URL-handling behavior.
    
### Checklist
    
    - [X] [Sign Allure CLA][cla]
    - [x] Provide unit tests
    

hasanalpzengin:
implements -> https://github.com/allure-framework/allure3/issues/822

epszaw (approved):

hasanalpzengin:
Thank you for the approval @epszaw ! Could you take a look at the pipeline? The failure doesn't seem to be related to my changes. It looks like the Allure Quality Gate step is failing with a 403 - Resource not accessible by integration, which suggests a permissions/integration issue rather than something introduced in this PR. I will be happy to take a look if it's something I can fix.

epszaw:
Hey, @hasanalpzengin it's ok! It doesn't relate to the PR directly, so we can merge the changes because the main checks have passed.
Thank you for the contribution!

Issue #820, opened 2026-07-24
History data points are always written with empty `url` — `executor.json` `reportUrl` is ignored (history links regression from Allure 2)


alex-opeykin:
### What happened?

Because `url` is empty, `TrHistoryItem` in the Awesome report renders the History tab entries as plain non-clickable rows. In Allure 2 the very same setup produced clickable history items out of the box: `HistoryPlugin` took `executorInfo.reportUrl` from `executor.json` and stored `<reportUrl>/#testresult/<uid>` per history item.

We are migrating our UI-test reporting from Allure 2 to Allure 3 (reports are built on Jenkins by the Jenkins Allure plugin, which drops `executor.json` with `buildUrl`/`reportUrl` into the results directory).

History itself works (`historyPath` + `appendHistory`): `history.jsonl` grows with every run. However, every history data point — and every entry inside its `testResults` — is written with `"url": ""`, even though `executor.json` with a valid `reportUrl` is present in the results.

### How can we reproduce it?

Repro attached (`history_empty_url.zip`): a results directory produced by the official NUnit starter pack (Allure.NUnit 2.15.0, 5 tests: 2 passed / 2 failed / 1 broken) plus `executor.json` exactly like the Jenkins Allure plugin writes it, and a minimal config.

`allure-results-1/executor.json`:

```json
{
  "name": "Jenkins",
  "type": "jenkins",
  "url": "http://127.0.0.1:58888/",
  "buildOrder": 1,
  "buildName": "demo #1",
  "buildUrl": "http://127.0.0.1:58888/",
  "reportUrl": "http://127.0.0.1:58888",
  "reportName": "Allure Report"
}
```

`allurerc.json`:

```json
{
  "name": "History empty url repro",
  "historyPath": "./history.jsonl",
  "appendHistory": true
}
```

Steps:

```bash
allure generate
cat history.jsonl
```

The appended data point has `"url": ""` at the top level and `"url": ""` in every `testResults` entry (the executor info is at the same time correctly shown in the generated report UI, so the file is definitely read).

```json
{"uuid":"...","name":"History empty url repro","timestamp":...,"knownTestCaseIds":[...],"testResults":{"<historyId>":{"id":"...","name":"AdditionWorks",...,"url":"","reportLinks":[]}, ...},"metrics":{},"url":""}
```

Run it a second time with another results directory — the History tab of the new report shows the previous run, but the entry is not a link.

[...]

PR #831, opened 2026-07-28
resolve history report links from executor metadata

### LINKED ISSUE #821: History tab links are missing a trailing slash — broken on hostings without directory redirects
### What happened?

Once the history point `url` is filled in (we had to inject it manually because of #820), the History tab entries become links — but the links are built unreliably.

`TrHistoryItem` (`packages/web-awesome/src/components/TestResult/TrHistory/TrHistoryItem.tsx`):

```ts
const { origin, pathname } = new URL(url);
const navUrl = new URL([pathname, reportOptions.id].join("/"), origin);
navUrl.hash = id;
```

**Primary problem — no trailing slash.** For a report URL with a path (typical CI case, e.g. Jenkins: `https://jenkins/job/X/42/allure`) the produced link is `…/allure/awesome#<id>` — a *directory* URL without the trailing slash. Whether it works now depends entirely on the hosting:

- `@allurereport/static-server` rescues it with a `301` to `…/awesome/` — the code even carries the comment `"redirect to directory with trailing slash to avoid invalid assets paths issue"`;
- the Jenkins Allure plugin (2.35.x), which serves reports straight from the archived `allure-report.zip`, performs **no** such redirect: it returns the *content* of `awesome/index.html` at the slash-less URL, all relative assets (`app-*.js`, `data/*`) resolve against the parent directory, and the page renders blank (`X-Content-Type-Options: nosniff` blocks the HTML fallbacks). This is our production case.

### How can we reproduce it?

Repro attached ([history_link_404.zip](https://github.com/user-attachments/files/30349772/history_link_404.zip)): two result sets from the official NUnit starter pack (same 5 tests / same history ids, `executor.json` like the Jenkins Allure plugin writes), a config with explicit `awesome` + `dashboard` plugins (so the report has the `awesome/` subdirectory — same layout as CI), and a helper script for the #820 workaround.

To give the report URL a path (like on Jenkins), the "previous build" report is generated into a subfolder and its *parent* is served:

```bash
allure generate allure-results-1 -o serve-old/build-1 --config allurerc.json

# workaround for #820: history url is never populated, inject it manually
# (sets url = http://127.0.0.1:58888/build-1)
node inject-url.mjs

allure generate allure-results-2 -o report-new --config allurerc.json

allure open serve-old  --port 58888    # terminal 1: hosting of the "previous build"
allure open report-new --port 58889    # terminal 2: the "current build" report
```

Open `http://127.0.0.1:58889/awesome/`, click any test (e.g. `BrokenByUnexpectedException`), switch to the **History** tab. The previous run's entry is a link with:

```
http://127.0.0.1:58888/build-1/awesome#5bd0de6d8fe94b75be93ae8ee778dd9e
```

— no trailing slash before the hash. It survives on your own static server *only* because of the rescue redirect:

```bash
curl -i http://127.0.0.1:58888/build-1/awesome
# HTTP/1.1 301 Moved Permanently
# Location: /build-1/awesome/
```

On a hosting without that redirect (Jenkins Allure plugin) the very same link loads `awesome/index.html` with broken relative assets → blank page.

Bonus (root-hosted manifestation): inject `http://127.0.0.1:58888` instead (no path) — the rendered link becomes `http://awesome/#<id>` and the click fails with `ERR_NAME_NOT_RESOLVED`, while `http://127.0.0.1:58888/awesome/#<id>` typed manually works fine.

### What did you expect?

The link ends with a trailing slash before the hash:

```
http://127.0.0.1:58888/build-1/awesome/#<id>
```

so that it does not depend on server-side directory redirects.

Better yet: resolve the final clickable URL at **generation time** — the generator knows its own layout (root vs `<pluginId>/` subdirectory) — and store it in the history point, instead of reconstructing it in the UI from the *current* report's plugin id (which also mangles root-hosted URLs into `http://<pluginId>/`).

### Environment

- Allure 3 version: 3.14.3
- Node.js version: 24.18.0
- Package manager and version: 11.16.0
- OS: Windows 10 Pro 22H2
- Jenkins Allure Plugin 2.35.2

### Additional context

- Related: #820 (history `url` is never populated at all; this issue is about how the link is built once `url` is there).
- Production impact: our reports are served by the Jenkins Allure plugin (2.35.x) at `https://jenkins/job/<job>/<build>/allure`. Every History link points to `…/allure/awesome#<id>` and opens a blank page, because the plugin serves the archived report without directory redirects. A one-character fix (trailing slash) on the allure side would make the links independent of the hosting behavior.

Image
Image
Image
Image
Image

formaceft-93:
fixes https://github.com/allure-framework/allure3/issues/821 https://github.com/allure-framework/allure3/issues/820 

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/web-awesome/src/components/TestResult/historyNavigation.ts
+++ b/packages/web-awesome/src/components/TestResult/historyNavigation.ts
@@ -1,0 +1,16 @@
+export const getHistoryNavigationUrl = (url: string | undefined, testResultId: string): string | undefined => {
+  if (!url) {
+    return undefined;
+  }
+
+  try {
+    const navUrl = new URL(url);
+
+    navUrl.pathname = navUrl.pathname.endsWith("/") ? navUrl.pathname : `${navUrl.pathname}/`;
+    navUrl.hash = testResultId;
+
+    return navUrl.toString();
+  } catch {
+    return undefined;
+  }
+};
--- a/packages/core/src/report.ts
+++ b/packages/core/src/report.ts
@@ -109,6 +109,16 @@ const remoteReportParams = (ci: CiDescriptor | undefined): { repo?: string; bran
 
 const errorDetails = (err: unknown): string => (err instanceof Error ? (err.stack ?? err.message) : String(err));
 
+const getExecutorReportUrl = (executor: unknown): string | undefined => {
+  if (!executor || typeof executor !== "object" || !("reportUrl" in executor)) {
+    return undefined;
+  }
+
+  const { reportUrl } = executor as { reportUrl?: unknown };
+
+  return typeof reportUrl === "string" && reportUrl.length > 0 ? reportUrl : undefined;
+};
+
 const closeReadStream = async (stream: ReadStream): Promise<void> => {
   if (stream.closed) {
     return;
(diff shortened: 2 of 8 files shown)


packages/web-awesome/src/components/TestResult/historyNavigation.ts
```diff
@@ -0,0 +1,22 @@
...
+  if (!url) {
+    return undefined;
+  }
+
+  try {
+    const navUrl = new URL(url);
+    const pathname = navUrl.pathname.endsWith("/") ? navUrl.pathname : `${navUrl.pathname}/`;
+    const lastSegment = pathname.slice(0, -1).split("/").pop();
+
+    navUrl.pathname = lastSegment === reportId ? pathname : `${pathname}${reportId}/`;
```
epszaw: Newly populated executor links target a nonexistent directory with the default single-report configuration. Core moves the sole plugin's contents to the output root, but an executor URL ending in `/allure` becomes `/allure/awesome/#id`, producing a 404. Persist/use the actual plugin destination, omit the plugin directory for flattened output, and cover navigation against a generated single-plugin report.

epszaw (changes requested):

epszaw (approved):

PR #840, opened 2026-07-31
feat(web-awesome): add search to Tags filter dropdown

### LINKED ISSUE #839: Add text search to the Awesome report Tags dropdown

We use custom tags to associate automated test results with software requirements. A test may cover one or several requirements, and a complete CI report may contain hundreds or thousands of unique requirement tags.

A reviewer often starts with a specific requirement ID and needs to find every test result associated with it across multiple suites, platforms, and test types.

The current Tags dropdown shows the entire tag list without a search field. The reviewer must manually scroll through hundreds or thousands of requirement IDs to find the desired tag.

For example, a report may contain tags such as:

- `REQ-1234`
- `REQ-1235`
- `REQ-1236`

If a reviewer needs evidence for `REQ-1235`, locating it manually in a very large dropdown is impractical.

This affects workflows such as:

- Requirements traceability and compliance reviews
- Finding all results associated with one requirement
- Reports combining multiple platforms and test frameworks
- Tests associated with multiple requirements
- Automatically generated requirement tags that are not suitable for manual browsing

The existing tag filter works correctly after a tag is selected. The problem is finding the desired tag in a large dropdown.

Add a text search field at the top of the Tags dropdown.

The user should be able to enter a complete or partial requirement ID, and the dropdown should display only matching tags. For example, entering `REQ-1235` should immediately narrow the list to that requirement.

Expected behavior:

- Search is case-insensitive.
- Full and partial tag values are supported.
- Clearing the search restores the complete tag list.
- Existing selections remain selected when the search text changes.
- Multiple tags can still be selected.
- Tests associated with multiple requirements continue to appear for each matching tag.
- Existing URL persistence and shareable filtered links remain unchanged.
- Keyboard navigation works within the filtered results.

This would make the existing tag-filter feature practical for large requirements-based test reports without changing the underlying result format or filtering behavior.

### Alternatives considered

_No response_

### Additional context

_No response_

MJayyusi:
### Context

The Awesome Tags dropdown lists every tag with no search field, which is impractical when a report has hundreds or thousands of requirement tags (e.g. `REQ-1235`).

This adds a case-insensitive partial-match search field at the top of the Tags dropdown so reviewers can find tags quickly. Existing multi-select, filtering, and URL persistence are unchanged.

Fixes #839

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


MJayyusi:
First contribution — please approve workflows when you have a moment.

MJayyusi:
@epszaw Can you please take a look or direct me on whats needed to get this one merged, this should be a simple one. 
Thank you
MJ

todti:
Thanks for the contribution, and sorry for the slow response here.

The approach looks right — reusing `SearchBox` and gating it behind a `searchable` prop on `ArrayFieldFilter` keeps the change small and leaves the other filters untouched. A few things before this can go in:

**1. Search query isn't reset when the dropdown closes**

`search` lives in `ArrayFieldFilter`, which stays mounted — only the `Menu` children unmount. So after typing `REQ-12` and closing the dropdown, reopening it still shows the filtered subset (and the same applies after the filter's Clear button). Worth resetting the query when the menu closes, and inside `onClear`.

**2. `changeDebounce={0}`**

This overrides the 300ms default, so every keystroke re-renders the whole option list synchronously. The list isn't virtualized, so on the thousands-of-tags reports this PR targets that's noticeable input lag. I'd just drop the prop and keep the default.

**3. Empty state**

When nothing matches, `Menu.Section` renders an empty `<ul>` — a blank dropdown that reads as broken. A short "nothing found" row would help.

**4. Dead `stopPropagation`**

The `onClick`/`onMouseDown` handlers on the search wrapper aren't needed: `Menu` only closes on clicks whose target is outside `menuRef`, and this div is inside the portal subtree (`Menu/index.tsx`, the `handleClickOutside` guard). They also swallow the click for every other document-level listener, so better removed.

**5. Rebase**

The branch conflicts with `main` — it's trivial, both sides appended a block at the end of `ReportFilters/styles.scss`. `ArrayFieldFilter` also picked up new consumers in the meantime (Severity, Resolution filters), worth a quick check that they still render as before.

A unit test covering the filtering (match, no-match, reset on close) would be good to have too.

MJayyusi:
@todti 
Thanks for the review - addressed in the latest push:

1. Search resets when the dropdown closes and when Clear is used
2. Dropped `changeDebounce={0}` (keeps the default debounce)
3. Added a “No results” empty state for non-matching search
4. Removed the unnecessary stopPropagation handlers
5. Rebased onto latest `main` (styles conflict resolved). Severity/Resolution still use `MultipleChoiceFieldFilter` and are unchanged
6. Added unit tests for match, no-match empty state, and reset on close

Ready for another look when you have a moment.

vdvukhzhilov (approved):
LGTM. Thanks!

PR #850, opened 2026-08-05
Enrich quality gate results

epszaw:
#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/core/src/qualityGate/qualityGate.ts
+++ b/packages/core/src/qualityGate/qualityGate.ts
@@ -40,7 +40,7 @@ export const convertQualityGateResultsToTestErrors = (results: QualityGateValida
 
 export class QualityGateState {
   #state: Record<string, { result: any; testResults: string[] }> = {};
-  #seenTestResultIds = new Set<string>();
+  #processedTestResultIds = new Set<string>();
 
   setResult(rule: string, value: any, testResults: string[] = []) {
     const previousTestResults = this.#state[rule]?.testResults ?? [];
@@ -59,14 +59,12 @@ export class QualityGateState {
     return [...(this.#state[rule]?.testResults ?? [])];
   }
 
-  getUnseenTestResults(testResults: TestResult[]) {
-    const uniqueTestResults = new Map(testResults.map((testResult) => [testResult.id, testResult]));
-
-    return [...uniqueTestResults.values()].filter(({ id }) => !this.#seenTestResultIds.has(id));
+  markTestResultsProcessed(testResults: TestResult[]) {
+    testResults.forEach(({ id }) => this.#processedTestResultIds.add(id));
   }
 
-  markTestResultsSeen(testResults: TestResult[]) {
-    testResults.forEach(({ id }) => this.#seenTestResultIds.add(id));
+  isTestResultProcessed(trId: string) {
+    return this.#processedTestResultIds.has(trId);
   }
 }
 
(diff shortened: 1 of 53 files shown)


packages/core/src/qualityGate/qualityGate.ts
```diff
@@ -39,14 +39,34 @@ export const convertQualityGateResultsToTestErrors = (results: QualityGateValida
...
+
+  getTestResults(rule: string) {
+    return [...(this.#state[rule]?.testResults ?? [])];
+  }
+
+  getUnseenTestResults(testResults: TestResult[]) {
+    const uniqueTestResults = new Map(testResults.map((testResult) => [testResult.id, testResult]));
+
+    return [...uniqueTestResults.values()].filter(({ id }) => !this.#seenTestResultIds.has(id));
+  }
```
vdvukhzhilov: This is only used to check if the TR is seen or not, so why not make it simpler?
_Besides its only input is `uniqueTrs` that are unique already._
```suggestion
  isUnseenTestResult(testResult: TestResult) {
    return !this.#seenTestResultIds.has(id);
  }
```



Then where it is used something like
```
    const trsToValidate = state?.isUnseenTestResult ? uniqueTrs.filter(state?.isUnseenTestResult) : uniqueTrs;
```


nit: `seen` -> `processed` and flip logic `isProcessedTestResult` ? 🤔

packages/plugin-awesome/src/generators.ts
```diff
@@ -620,8 +632,35 @@ export const generateGlobals = async (
...
+  } = {},
 ) => {
-  await writer.writeWidget("quality-gate.json", qualityGateResults);
+  const { tests = [], labels = [], appendTitlePath } = options;
+  const testsById = new Map(tests.map((test) => [test.id, test] as const));
+  const resultsWithTrees = Object.fromEntries(
+    Object.entries(qualityGateResults).map(([environment, results]) => [
+      environment,
+      results.map((result) => {
+        const relatedTests = [...new Set(result.testResults)]
```
formaceft-93: Custom rules or older persisted quality gate results may still not have 'testResults' field, and spreading 'undefined' will crash the report generation. May be we can use 'result.testResults ?? []' here to keep the new field backwards-compatible?

packages/core/src/qualityGate/rules.ts
```diff
@@ -70,17 +92,22 @@ export const maxDurationRule: QualityGateRule<number> = {
  * Fails if any test in the run does not have the given environment.
  * Expected: environment name (string).
  */
-export const allTestsContainEnvRule: QualityGateRule<string> = {
+export const allTestsContainEnvRule: QualityGateRule<string, number> = {
   rule: "allTestsContainEnv",
   message: ({ actual, expected }) =>
     `Not all tests contain the required environment, ${bold(`"${expected}"`)}; ${bold(actual.length === 1 ? "one" : String(actual))} ${actual.length === 1 ? "test has" : "tests have"} different or missing environment`,
```
formaceft-93: we still check here 'actual.length' however on 103 line we count actual as number 'testsWithoutEnv.length + previous'
so here better to use 'actual === 1' for correct message?

formaceft-93 (approved):

packages/core/src/qualityGate/qualityGate.ts
epszaw: 👍

packages/plugin-awesome/src/generators.ts
epszaw: 👍

packages/core/src/qualityGate/rules.ts
epszaw: Simplified the message 👍

PR #851, opened 2026-08-06
known issues ui

formaceft-93:
- added a Known issues tree quick filter and tree marker icon
- added global and test-result **Known Issues** tabs
- added known issue links/details and test-result grouping
- excluded known issue failures from the success-rate chart

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


epszaw (changes requested):
Found some high-level issues:

1. The summary page widgets show the run is broken, but there's a known issue only, so we can't treat the report as broken. The chart on the page is a success rate chart:Image

Let's also render the known issue count in the widgets.

2. I recommend using the same approach for links rendering as in the report's UI:

**Example with test result links (issue, tms)**:

Image

**Current known issue links (issue, tms)**:
Image

3. The tooltip should be just `Known issue`:

Image

4. Let's use the same test result row component in the known issues sub-tree with more data, such as duration, etc.:

**Main tree**:

Image

**Known issue sub-tree**:

Image

PR #855, opened 2026-08-08
fix(directory-watcher): find allure-results in dot-prefixed dirs

meganemura:
## Summary

- `findMatching` (used by `allure agent -- <cmd>` and `allure run -- <cmd>` to
  watch for `allure-results` directories while the test command runs)
  skipped every directory whose name starts with `.`. Results written under
  a dot-prefixed path, e.g. `.reports/allure-results/`, were silently
  missed, and the run was reported as `no tests` with a `no-tests-observed`
  finding.
- Other commands (`allure watch`, `allure agent inspect`, and the one-shot
  generators like `allure awesome`) resolve `allure-results` through a
  different, glob-based lookup that already passes `dot: true`, so they
  never had this problem.
- Fix: only skip `.git` and `node_modules` by name (both large, and never
  contain `allure-results`); stop skipping directories just for starting
  with `.`.

## Test plan

- `yarn vitest run` in `packages/directory-watcher` — 26/26 pass (new/
  updated cases: dot-prefixed discovery, `.git`/`node_modules` still
  skipped, nested dot-prefixed directories)
- `tsc --noEmit` — clean

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/directory-watcher/src/watcher.ts
+++ b/packages/directory-watcher/src/watcher.ts
@@ -40,8 +40,10 @@ export const findMatching = async (
     for await (const dirent of dir) {
       const path = join(dirent.parentPath ?? dirent.path, dirent.name);
 
-      // shouldn't be looking in private folders
-      if (dirent.name.at(0) === "." || dirent.name === "node_modules") {
+      // .git and node_modules are always skipped: this watcher re-walks repeatedly on a
+      // short interval for the life of the run, unlike a one-off glob call, so recursing
+      // into them on every pass is wasteful.
+      if (dirent.name === ".git" || dirent.name === "node_modules") {
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


vdvukhzhilov:
@meganemura Thank you for the PR and for describing the issue 🎉 

[Latest Allure 3 release](https://github.com/allure-framework/allure3/releases/tag/v3.16.0) brings a way to tell Allure exactly where to look for results: `resultsDir` in the `allurerc` config, and `--results-dir` for `allure run` / `allure agent` commands.

For your case (results under something like `.reports/allure-results`), that should be just the right thing - you point Allure at that path (or a pattern that includes it), and those results are picked up during the run.

Example config:

```js
export default {
  resultsDir: ".reports/allure-results",
  // ...
};
```

Or on the command line:

```shell
allure run --results-dir .reports/allure-results -- <your test command>
```

So if you’re fine setting that once, you won’t need a watcher change for this scenario. 
If you’d still prefer zero-config discovery of results under any `.…` folder, we can keep discussing this PR 👋

meganemura:
Thanks for adding `resultsDir` and `--results-dir`. That is enough for my case, so I will close this.

If this still causes trouble for someone using zero-config discovery, it may be worth looking at again.

PR #861, opened 2026-08-11
fix(core): keep allure serve alive when historyPath is configured

### LINKED ISSUE #827: [CLI] `allure serve` silently exits without starting the server when `historyPath` is set
Title: [CLI] `allure serve` silently exits without starting the server when `historyPath` is set

### Describe the bug

When `allurerc.mjs` contains a `historyPath`, `allure serve` generates the report (plugins run, history file is written) but then **exits silently without starting the HTTP server** — no "Allure is running on ..." line, no error, nothing to connect to. Removing `historyPath` from the config makes `serve` work again.

`allure generate` + `allure open` work fine with the same config, so the history feature itself is OK — only the `serve` flow is affected.

### To Reproduce

```bash
mkdir -p repro/allure-results && cd repro

cat > allure-results/x-result.json <<'EOF'
{"uuid":"11111111-1111-1111-1111-111111111111","historyId":"abc","name":"dummy test","status":"passed","stage":"finished","start":1784900000000,"stop":1784900001000,"labels":[],"steps":[],"parameters":[],"links":[]}
EOF

cat > allurerc.mjs <<'EOF'
export default {
    name: 'repro',
    historyPath: './allure-history.jsonl',
};
EOF

npx allure serve allure-results
# -> generates, writes allure-history.jsonl, then exits silently. No server, no output.
```

Now remove `historyPath`:

```bash
echo "export default {name: 'repro'}" > allurerc.mjs
npx allure serve allure-results
# -> "Allure is running on http://localhost:<port>" — works.
```

### Expected behavior

`allure serve` starts the server (and prints the URL) regardless of whether run history is configured — or at least fails with an explicit error message.

### Environment

- allure (npm CLI): 3.14.3
- Node.js: 24.18.0
- OS: macOS 15 (darwin 25.5.0)
- Reproduces both with a single default plugin and with multiple plugins (awesome/dashboard/log); bisected — `historyPath` is the only trigger.

svasenkov:
## Summary

- Fixes `allure serve` silently exiting after report generation when `historyPath` is set in `allurerc.mjs` (Fixes #827).
- Root cause: `AllureLocalHistory.appendHistory()` piped into a `FileHandle` write stream with `autoClose: false` and closed the handle before the stream was destroyed. On Node.js 20+ this leaves a pending promise and can terminate the process before `serve` starts.
- Finish and destroy the write stream before closing the file handle.

## Test plan

- [x] `packages/core` history tests, including a standalone Node subprocess reproducing the FileHandle stream lifecycle
- [x] `packages/cli` integration test: `allure serve allure-results` with `historyPath` prints `Allure is running on http://localhost:…` and keeps running
- [x] Manual repro against published `allure@3.14.3` flow (history written, process used to exit with code 0 before server URL)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/cli/test/commands/open.integration.test.ts
+++ b/packages/cli/test/commands/open.integration.test.ts
@@ -1,0 +1,153 @@
+import { spawn } from "node:child_process";
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { dirname, join, resolve } from "node:path";
+import process from "node:process";
+import { fileURLToPath } from "node:url";
+
+import { epic, feature, label, story } from "allure-js-commons";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+const commandsDir = dirname(fileURLToPath(import.meta.url));
+const repoRoot = resolve(commandsDir, "../../../..");
+const cliPath = join(repoRoot, "packages", "cli", "cli.js");
+const yarnRcPath = join(repoRoot, ".yarnrc.yml");
+
+const workspaces = new Set<string>();
+
+const resolveYarnInvocation = async () => {
+  const { readFile, stat } = await import("node:fs/promises");
+  const yarnRc = await readFile(yarnRcPath, "utf-8");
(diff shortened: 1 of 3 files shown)


andrcuns:
Superseded by: https://github.com/allure-framework/allure3/pull/962

PR #876, opened 2026-08-19
chore: add budget checks

vdvukhzhilov:
#### Checklist

- [x] [Sign Allure CLA][cla]
- [ ] Provide unit tests


epszaw (changes requested):
Why can't we use https://www.npmjs.com/package/size-limit + allure external check?

vdvukhzhilov:
> https://www.npmjs.com/package/size-limit

I don't want to bring another dependency 😆 

Also it does not cover our use cases fully

epszaw (approved):

PR #882, opened 2026-08-21
Add `resultsDir` config option and `--results-dir` run command arg for scoped Allure results discovery

vdvukhzhilov:
### Context

Introduce `resultsDir` so users can choose which Allure results directories feed reports and live runs.

- **Config:** `resultsDir` (`string | string[]`) in `allurerc`
- **CLI:** `--results-dir` on `run` / `agent` (repeatable);
- **Priority:** CLI argument > config option > command default value

Main reason: avoid pulling every `allure-results` tree in the repo into `allure run` reports when only a subset is relevant.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/core/src/utils/resultsDir.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Drops "", whitespace-only entries, and empty arrays — treated as unset → [].
+ */
+export const normalizeResultsDir = (value?: string | string[]): string[] => {
+  if (value === undefined) {
+    return [];
+  }
+
+  const values = Array.isArray(value) ? value : [value];
+
+  return values.map((entry) => entry.trim()).filter((entry) => entry.length > 0);
```
delatrie: I would not trim file paths. Whitespaces are significant in this context.

packages/cli/src/utils/resultsPatterns.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Drops "", whitespace-only entries, and empty arrays — treated as unset → [].
+ */
+export const normalizeResultsDir = (value?: string | readonly string[]): string[] => {
+  if (value === undefined) {
+    return [];
+  }
+
+  const values = typeof value === "string" ? [value] : [...value];
+
+  return values.map((entry) => entry.trim()).filter((entry) => entry.length > 0);
```
delatrie: Same here: a leading/trailing whitespace is significant in paths (although weird).

packages/cli/src/commands/history.ts
```diff
@@ -47,7 +47,14 @@ export class HistoryCommand extends Command {
   });
 
   async execute() {
-    const { resultDirectories, patterns } = await findAllureResultDirectories(process.cwd(), this.resultsDir);
+    const cwd = process.cwd();
+    const fileConfig = await readConfig(cwd);
```
formaceft-93: This reads the full resolved config only to access `resultsDir`, but `readConfig()` also resolves and instantiates configured plugins. For commands like `history`, `testplan`, and `results pack`, plugin resolution is unrelated and can now make the command fail when the project has a custom/missing/side-effectful report plugin, even if the user passed explicit result directories.

Could we use `readRawConfig()` here and normalize only `rawConfig.resultsDir` for discovery? That preserves the new `resultsDir` behavior without making these utility commands depend on report plugin loading.

formaceft-93:
`allure agent` now supports repeatable `--results-dir`, but the capability payload does not advertise it in `commands.run.options`. Agents/tooling use `allure agent capabilities --json` to decide which flags are safe to pass, so they may miss the new scoped discovery option.

Please add `"--results-dir"` to the run options here and cover it in `packages/plugin-agent/test/capabilities.test.ts`.

formaceft-93 (approved):

packages/cli/src/commands/commons/resultsDiscovery.ts
```diff
@@ -0,0 +1,34 @@
...
+  const { abortController = new AbortController(), indexDelay = 600 } = options;
+  let previousAllureResults: Set<string> = new Set();
+
+  const callback = async () => {
+    if (abortController.signal.aborted) {
+      return;
+    }
+
+    const matched = await findDirectoriesByGlobs(cwd, patterns);
+    const currentAllureResults = new Set(matched.map(normalizeResultsDirectoryPath));
```
delatrie: Is there any point in normalizing? `findDirectoriesByGlobs`delegates to `glob` with `{ mark: true, nodir: false, absolute: true }` then filters directory matches, which means every returned entry is an absolute path to a directory.

Since all entries returned by`findDirectoriesByGlobs` are absolute, have trailing slashes, and no entries come from other sources, they should be normalized enough for the difference check without `normalizeResultsDirectoryPath` (which just removes trailing slashes; its `resolve` does not change the path as it's already absolute).

packages/cli/src/commands/results/pack.ts
```diff
@@ -27,13 +29,17 @@ export class ResultsPackCommand extends Command {
 
   resultsDir = Option.String({
     required: false,
-    name: "Pattern to match test results directories in the current working directory (default: ./**/allure-results)",
+    name: "Pattern to match test results directories (CLI > config.resultsDir; when both empty: ./**/allure-results)",
```
delatrie: Here and for other commands, I suggest something less cryptic, e.g.:

```suggestion
    name: "Pattern to match test results directories. Overrides config.resultsDir. Defaults to ./**/allure-results when neither is set.",
```

packages/cli/src/commands/results/pack.ts
```diff
@@ -27,13 +29,17 @@ export class ResultsPackCommand extends Command {
...
   });
 
   name = Option.String("--name", {
     description: "The archive name (default: allure-results.zip)",
   });
 
+  config = Option.String("--config,-c", {
+    description: "The path Allure config file",
```
delatrie:
```suggestion
    description: "The path to Allure config file",
```

delatrie (approved):
There are some minor issues, it's up to you whether and how to fix them. Other than that, looks good!

PR #886, opened 2026-08-24
Resolutions

epszaw:
### Context

The new Resolutions feature lets users classify failed and broken tests with configurable matching rules. Rules can target error messages, test case IDs, retry hashes, or environments.

Configure resolution categories in `allurerc.mjs`:

```js
export default {
  resolutions: {
    knownIssuesPath: "./known-issues.json",
    links: {
      jira: {
        urlTemplate: "https://jira.example.com/browse/%s",
        nameTemplate: "Issue %s",
      },
    },
    rules: [
      {
        resolution: "issue",
        testCaseId: ["checkout-test-id"],
        issue: {
          id: "SHOP-123",
          type: "jira",
        },
        comment: "Tracked by the checkout team",
      },
      {
        resolution: "muted",
        retryHash: ["unstable-retry-hash"],
        comment: "Ignore this known infrastructure problem",
      },
      {
        resolution: "accepted",
        messageRegexp: "Unsupported browser version",
        environment: ["legacy-browser"],
        comment: "We accept this risk for the legacy environment",
      },
    ],
  },
};
```

Available resolution categories are:
- `issue` — use when the failure has an issue in an issue tracker that can be linked to the test result.
- `muted` — use for a problem that should be ignored.
- `accepted` — use when the failure represents a known risk that the team accepts.

Muted and accepted failures do not block quality gates, while linked issues remain visible as failures. Issue resolutions can also be exported to the configured known-issues file. The test result page lets users copy the full name, test case ID, or retry hash needed to create precise rules.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/plugin-awesome/src/converters.ts
```diff
@@ -45,7 +45,6 @@ export const convertTestResult = (
     historyId: tr.historyId,
     flaky: tr.flaky,
     muted: tr.muted,
-    known: tr.known,
```
formaceft-93, line 48: Should we have to use new resolution field here or it will be my responsibility in next ui tasks? (with passing them to TrInfo component)

formaceft-93:
Reader-api still has have _known_ field [here](https://github.com/allure-framework/allure3/blob/main/packages/reader-api/src/model.ts)

packages/cli/src/commands/generate.ts
formaceft-93, line 79: This option now sounds more like it reads known issues from a file, but the code only writes the generated snapshot there. Could we make the name/description explicit, e.g. “known issues output path”, unless reading this file is still intended?

packages/core/src/resolutions.ts
```diff
@@ -0,0 +1,253 @@
...
+  config: ResolutionsConfig,
+  issueIds: Set<string>,
+  errors: string[],
+) => {
+  if (!nonEmpty(rule.issue?.id) || !nonEmpty(rule.issue?.type)) {
+    errors.push(`${path}.issue.id and ${path}.issue.type must be non-empty strings`);
+    return;
+  }
+
+  if (issueIds.has(rule.issue.id)) {
```
formaceft-93, line 111: This makes issue.id unique across all rules, so one real issue can’t be matched by multiple independent matcher rules. Arrays cover simple cases like several testCaseIds, but not different matcher shapes, because fields inside one rule are combined with AND. Is this restriction intentional? If yes, maybe worth documenting; if not, we may need to separate issue metadata from matchers, or allow repeated issue IDs and merge them in the report.

formaceft-93 (approved):

packages/cli/src/commands/generate.ts
epszaw: Updated the description

packages/plugin-awesome/src/converters.ts
epszaw: Yes, we need to utilize the `resolution_category` field further

packages/core/src/resolutions.ts
epszaw: Added the resolutions categories' priority. There can be only a single resolution category, but multiple tests can be related to multiple issues at the same time with intersection

packages/sandbox/known-issues.json
```diff
@@ -0,0 +1 @@
...runner/dist/chunk-artifact.js:302:11\n    at file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/cache/@vitest-runner-npm-4.1.8-38f4df6de4-278d148212.zip/node_modules/@vitest/runner/dist/chunk-artifact.js:1903:26\n    at file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/cache/@vitest-runner-npm-4.1.8-38f4df6de4-278d148212.zip/node_modules/@vitest/runner/dist/chunk-artifact.js:2326:20\n    at new Promise (<anonymous>)\n    at runWithCancel (file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/cache/@vitest-runner-npm-4.1.8-38f4df6de4-278d148212.zip/node_modules/@vitest/runner/dist/chunk-artifact.js:2323:10)\n    at file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/cache/@vitest-runner-npm-4.1.8-38f4df6de4-278d148212.zip/node_modules/@vitest/runner/dist/chunk-artifact.js:2305:20\n    at new Promise (<anonymous>)\n    at runWithTimeout (file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/cache/@vitest-runner-npm-4.1.8-38f4df6de4-278d148212.zip/node_modules/@vitest/runner/dist/chunk-artifact.js:2272:10)\n    at file:///Users/epszaw/Code/Work/qameta/allure3/.yarn/cache/@vitest-runner-npm-4.1.8-38f4df6de4-278d148212.zip/node_modules/@vitest/runner/dist/chunk-artifact.js:2955:64"}}}}]}
```
formaceft-93, line 1: local machine files, is okay?

packages/core/src/store/store.ts
```diff
@@ -466,32 +460,53 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
+    this.#removeResolutionIssueAssociation(testResult.id);
 
-    if (historyIdCandidates.length === 0) {
-      return false;
+    if (rule?.resolution !== "issue") {
+      return;
     }
 
-    const ruleKnownIssue = getKnownIssueByRules(testResult, this.#knownIssuesConfig, environmentId);
+    const resolutionIssue: ResolutionIssue = { ...rule.issue, comment: rule.comment };
```
formaceft-93, line 474: We require and validate resolutions.links, but the matched issue stored here only keeps id, type, and comment. I couldn’t find where urlTemplate / nameTemplate becomes an actual link. Should we materialize the link here, or keep links optional until it is used?

formaceft-93 (approved):

packages/core/src/store/store.ts
eddy-seager: I would like to use the links. Is there an issue to track this?

PR #888, opened 2026-08-25
fix(plugin-testops): Send the CI descriptor TestOps actually expects

todti:
### Context

Uploading results from an Azure DevOps pipeline produces a Job in TestOps that

1. is named `<projectGuid>_<definitionId>` instead of the pipeline name, and
2. cannot be triggered back from TestOps — the run fails with "build server is not configured for job".

These look like one bug but are two independent ones.

**The job is never bound to a build server.** `startUpload()` sent the CI block as `{ name: ci.type }`, while TestOps identifies the CI by `type` and `endpoint` and ignores anything else. With neither of them present it cannot tell which build server integration the job belongs to, so the job is stored without one — and the unknown property is dropped without an error, so the upload succeeds and nothing hints at the problem until someone presses Run. This is not Azure-specific: the endpoint was never sent for any CI, so build server binding never worked anywhere.

**The job name was the uid.** `job.name` was set to `ci.jobUid`, and `ci.jobName` (`Build.DefinitionName` on Azure, `Release.DefinitionName` for classic releases) was never sent. Triggering itself relies on `job.uid`, which already has the shape TestOps parses — so the name is a display defect rather than the reason runs fail.

### What changed

- `ci: { type, endpoint }` instead of `ci: { name }`. The endpoint is the origin of `ci.jobUrl` — enough for TestOps to recognise the integration, and unlike `SYSTEM_COLLECTIONURI` it carries no trailing slash.
- `job.name` falls back to the uid only when the CI reports no name, and `job.url` is sent.
- `jobRun.name` and `jobRun.url` are sent; TestOps stores both and refreshes them on every upload.
- Azure job and job run URLs no longer carry a double slash: `SYSTEM_COLLECTIONURI` always ends with one, and the URLs were built by concatenation. Harmless while nothing read them, but the job URL is stored once and never refreshed.
- `circle` and `amazon` are reported under the names TestOps uses for those integrations (`circleci`, `aws`). The job uids both detectors produce already match what it expects, so the type name was the only thing missing.

### Known limitations

- Not retroactive. TestOps keeps the name and the integration of a job it already knows, so jobs created before this fix keep their name and stay unbound — they have to be re-created there.
- `drone` still resolves to no integration type: TestOps has none matching it.

#### Checklist

- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests


todti:

PR #889, opened 2026-08-25
fix(plugin-testops): Align the upload payload with the backend contract

todti:
### Context

Three defects found by comparing what the plugin uploads against what the TestOps API accepts. They are independent of each other and of #888, which fixes the `ci` block of the same request.

**A blank error message rejects the whole batch.** Global errors come from the results files test framework adapters write, and their message is optional in our model — nothing validates it on the way in. TestOps requires a non-blank message and rejects the entire bulk request when one item lacks it, so a single untitled error dropped every valid error uploaded alongside it. A placeholder is sent instead, keeping the trace. This became easier to hit once every not-yet-uploaded error started being sent rather than only the last one.

**A long launch name fails launch creation.** TestOps accepts a launch name of at most 255 characters and fails the request otherwise, so a longer report title killed the upload before it started. The name is trimmed to the accepted length now.

**`environment` was sent into the void.** The upload API has no `environment` property on a test result, so the value was dropped on arrival; the binding it was meant to carry travels in `namedEnv`, which is sent next to it. The dead field is gone from the payload and from the type that mirrors the API.

#### Checklist

- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests

### DIFF
--- a/packages/plugin-testops/src/client.ts
+++ b/packages/plugin-testops/src/client.ts
@@ -55,6 +55,8 @@ class TestOpsClientError extends AxiosError<{
   request: ClientRequest;
 }
 
+const UNKNOWN_ERROR_MESSAGE = "Unknown error";
+const MAX_LAUNCH_NAME_LENGTH = 255;
 const CHUNK_SIZE = 100;
 const BULK_UPLOAD_CHUNK_SIZE = 1000;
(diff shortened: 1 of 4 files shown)


todti:

PR #903, opened 2026-08-31
Replace history id by retry hash

epszaw:
#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/core/src/store/convert.ts
```diff
@@ -121,22 +129,29 @@ export const testResultRawToState = (stateData: StateData, raw: RawTestResult, c
...
+    const maybeTestCase = testCases.get(testCaseHash);
+
     if (maybeTestCase) {
       return maybeTestCase;
     }
 
     const testCase: TestCase = {
-      id,
-      allureId,
+      id: testCaseHash,
```
formaceft-93, line 146: The spec describes `_fallbackTestCaseId` as a migration path when an adapter changes its test id: current `testCaseHash` may be new, but fallback hash should let us find the old test case/history. In this PR `TestCase.id` is set to the current `testCaseHash`, so when the adapter id changes, the product-level test case id changes too. Where do we use `_fallbackTestCaseId` to attach the new hash back to the old test case/history, or is that intentionally deferred?

packages/core/src/store/store.ts
```diff
@@ -1237,13 +1259,11 @@ export class DefaultAllureStore implements AllureStore, ResultsVisitor {
...
 
-    const historyIdCandidates = getHistoryIdCandidates(tr);
-
-    if (historyIdCandidates.length === 0) {
+    if (!tr.retryHash) {
       return [];
     }
 
-    return selectHistoryTestResults(this.#historyPoints, historyIdCandidates);
+    return selectHistoryTestResults(this.#historyPoints, [tr.retryHash]);
```
formaceft-93, line 1294: Can we clarify how `_fallbackTestCaseId` is supposed to affect history lookup? Right now history is selected only by the current `retryHash`, so if an adapter changes test id from A to B and sends `_fallbackTestCaseId=A`, the old history under A still won’t be found. Is fallback-based history migration intentionally deferred to a later catalog/index change?

packages/core/src/store/convert.ts
```diff
@@ -73,8 +77,10 @@ export const testResultRawToState = (stateData: StateData, raw: RawTestResult, c
   const hostId = findByLabelName(labels, "host");
   const threadId = findByLabelName(labels, "thread");
   const name = raw.name || "Unknown test";
-  const testCase = processTestCase(stateData, raw);
   const parameters = convertParameters(raw.parameters);
+  const testCaseHash = calculateTestCaseHash(raw.testId, raw.fullName);
```
formaceft-93, line 81: Current allure-js adapters often emit an already-hashed `testCaseId`. With this PR we always compute `testCaseHash = md5(raw.testId)`, so old allure-js output becomes double-hashed, while future canonical output likely won’t. Can we add old-vs-new allure-js fixtures or document the accepted mixed-version behavior?

packages/core/src/report.ts
```diff
@@ -798,9 +798,9 @@ export class AllureReport {
               const globalAttachmentsEntry = await requiredEntryData(AllureStoreDumpFiles.GlobalAttachments);
               const globalErrorsEntry = await requiredEntryData(AllureStoreDumpFiles.GlobalErrors);
               const indexAttachmentsEntry = await requiredEntryData(AllureStoreDumpFiles.IndexAttachmentsByTestResults);
-              const indexTestResultsByHistoryId = await requiredEntryData(
-                AllureStoreDumpFiles.IndexTestResultsByHistoryId,
-              );
+              const indexTestResultsByRetryHash =
```
formaceft-93, line 890: Can we clarify what compatibility this legacy history-id read gives us? Since restore rebuilds indexes from test results with the new `retryHash` rules, I’m not sure old history-id keys are actually migrated here?

packages/core-api/src/utils/identity.ts
```diff
@@ -0,0 +1,94 @@
...
+  const normalized: { name: string; value: string }[] = [];
+
+  for (const parameter of parameters ?? []) {
+    if (parameter == null) {
+      continue;
+    }
+
+    const { name } = parameter;
+
+    if (typeof name !== "string" || name.length === 0 || parameter.excluded === true) {
```
formaceft-93, line 57: We calculate `parametersHash` from raw params on first ingest, but after restore we calculate it from converted `TestResult.parameters`. Are those guaranteed to produce the same hash for null, hidden/masked, excluded, malformed, and duplicate params? A dump/restore regression test for these cases would help.

formaceft-93 (approved):

andrcuns:
I went over the code but with a very limited experience on working with this codebase can't really say I could thoroughly reviewed it, but once concern that seems to arise if I followed the change correctly, is history continuity after this change.

In `utils/history.ts` we basically reassign old `historyId` as `retryHash` which will be used for lookups if I followed the code correctly. Old historyId did not take environment in to account which new `retryHash` does, so it might end up creating a mismatch?

I believe similar issue could be with ALLURE_ID? It was used for old identity but not for new one?

packages/core/src/store/convert.ts
epszaw: This is intentional for the current storage mode.

The existing Allure 3 history is keyed by `retryHash`; it is not the cross-report test-case catalog/history index described in the identity specification. `TestCase.id` here is therefore a current-report grouping ID derived from `testCaseHash`, not a stable `catalogTestCaseId`.

Without a cross-report catalog, `_fallbackTestCaseId` must be ignored and an integration identity change is allowed to change `TestCase.id` and break history continuity.

When a file-based catalog is introduced, fallback resolution will belong there: it will find the existing catalog record by the fallback hash, preserve its stable `catalogTestCaseId`, and update only its stored `testCaseHash`. That catalog behavior is not being simulated in the current in-memory test-case map.

packages/core/src/store/store.ts
epszaw: Fallback-based history lookup is intentionally not performed here.

`_fallbackTestCaseId` is not an alternative history or retry key. Without a cross-report test-case catalog, history is keyed only by the current `retryHash`, and fallback labels must be ignored. Looking up history through a retry hash derived from the fallback identity would violate that rule.

Once a catalog/history index exists, the fallback will instead resolve the stable catalog test case and update its current `testCaseHash`. Cross-report history will then remain attached through `catalogTestCaseId + parametersHash + environmentHash`, rather than through a fallback-derived retry hash.

packages/core/src/store/convert.ts
epszaw: `testCaseId` is defined as an opaque integration-supplied value. Consumers must calculate `testCaseHash = md5Utf8(testCaseId)` over the exact supplied value, even when that value happens to look like an MD5 hash.

Detecting a 32-character hexadecimal value and treating it as a precomputed `testCaseHash` would reinterpret the adapter identity and make the calculation dependent on its shape.

If an integration changes from its previous identifier to a new identifier, that is an identity migration. With a cross-report catalog, it should emit the previous raw adapter value as `_fallbackTestCaseId`; without a catalog, loss of history continuity is expected. Therefore, we should not add an old-allure-js hashing special case in the consumer.

epszaw:
> I went over the code but with a very limited experience on working with this codebase can't really say I could thoroughly reviewed it, but once concern that seems to arise if I followed the change correctly, is history continuity after this change.
> 
> In `utils/history.ts` we basically reassign old `historyId` as `retryHash` which will be used for lookups if I followed the code correctly. Old historyId did not take environment in to account which new `retryHash` does, so it might end up creating a mismatch?
> 
> I believe similar issue could be with ALLURE_ID? It was used for old identity but not for new one?

Both differences are intentional under the new identity rules.

An explicitly resolved named environment participates in `retryHash`, while the default environment does not. This separates retries and history for named environments. Existing history keys that did not include the environment component cannot always be translated to the new key.

`ALLURE_ID` and `AS_ID` are external catalog references and must not participate in `testCaseHash`, `parametersHash`, `environmentHash`, or `retryHash`. The fact that the previous identity calculation could involve `ALLURE_ID` is behavior this change intentionally removes.

The current file history is retry-hash keyed and does not yet contain a cross-report test-case catalog. In that mode, identity changes may break continuity by design. `normalizeHistoryTestResults` can preserve old data only when its existing key already equals the newly calculated `retryHash`; it is not a general identity migration mechanism.

A future file-based test-case catalog can preserve continuity by resolving the stable catalog test case, including through `_fallbackTestCaseId`, and linking history through that stable catalog ID.

packages/core/src/report.ts
epszaw: The legacy history-id dump index has been removed.

Restore no longer treats old history-id keys as retry hashes. Serialized retry
and test-case identity indexes are ignored, canonical hashes are recalculated
from restored results, and both indexes are rebuilt from that canonical state.

Legacy project history compatibility is handled separately through read-only
history lookup and does not use the dump index.

packages/core-api/src/utils/identity.ts
epszaw: Initial ingestion now converts parameters before calculating parametersHash,
so ingestion and dump restoration hash the same stored representation.

The dump/restore regression matrix covers absent and empty parameters, malformed
or empty names, null values, excluded parameters, hidden and masked values,
duplicates, same-name/different-value pairs, utf8 ordering, and adapter-provided
hash overrides. parametersHash and retryHash remain unchanged after restoration.

formaceft-93 (approved):

vdvukhzhilov (approved):

PR #906, opened 2026-09-01
fix(plugin-testops): correlate CI uploads with TestOps job runs end-to-end

todti:
## Summary

Reworks `plugin-testops` to correlate CI-triggered uploads with an existing TestOps job run end-to-end, instead of always creating a new launch:

- Attaches uploads to an existing TestOps job run when one is provided, or to an existing launch via a `launchId` option, instead of always creating a new launch
- Adds CI-descriptor env overrides and an `ALLURE_CI_ENV` passthrough for pipeline steps that don't share env with the step that ran the plugin
- Adds an upload retry queue and rate pacer so a TestOps outage defers failed upload batches to finalization instead of dropping them
- Fixes the Azure CI detector's `getRootURL` to strip a trailing slash from `SYSTEM_COLLECTIONURI`
- Adds a `testops-plan` CLI command that fetches a job run's test plan from TestOps

## Fixes

- Job runs started directly by TestOps (not through a recognized CI provider) now correctly attach to their launch instead of crashing the run
- The job's branch and custom-name defaults no longer get overwritten with stale values on repeat runs of the same job
- Launch creation for local (non-CI) runs no longer fails, by sending only the fields the TestOps API accepts
- A bulk upload of global errors is no longer rejected outright when one error in the batch has a blank message
- Long report/launch names no longer fail upload; they're truncated to the accepted length instead
- Removed a redundant retry layer that could redo already-successful work during a TestOps outage
- The test-plan command now sees the job run id in pipeline setups where it's passed through an encoded environment bundle instead of a plain variable
- Stale queued upload entries no longer linger after a later successful retry, which could otherwise cause a duplicate re-upload at finalization

## Notes

Split the larger free functions out of `client.ts`/`plugin.ts` into `utils/` modules for readability, without changing the `TestOpsClient`/`TestOpsPlugin` class structure.


packages/plugin-testops/src/client.ts
```diff
@@ -233,30 +196,85 @@ export class TestOpsClient {
...
+          job: jobPayload,
+          jobRun: jobRunPayload,
+          ...(this.#launch ? { launch: { id: this.#launch.id } } : {}),
         },
-      },
-    });
+      }),
+    );
+
+    if (data?.jobRunId) {
```
formaceft-93: I think we can lose the known job run here. When startUpload() is called with an existing jobRunId, we only persist it if /api/upload/start returns data.jobRunId. But the response type makes jobRunId optional, and some tests already allow a response with only launchId. In that case createSession() falls back to a manual launch session, so uploaded results are no longer bound to the TestOps job run. Could we store the provided jobRunId as a fallback, e.g. data?.jobRunId ?? jobRunId, and add a regression test for the “launchId only” response?

packages/plugin-testops/src/plugin.ts
```diff
@@ -660,6 +637,19 @@ export class TestOpsPlugin implements Plugin {
...
     await this.#upload(store, { context, stage: "done" });
 
+    const stillPending = await this.#uploadQueue.flush({
+      onRetry: async (error, attempt) => {
+        this.#logger.debug(`Retrying deferred TestOps upload at finalization (attempt ${attempt}): ${error}`);
+        await this.#reopenLaunchIfClosed(error);
+      },
+    });
+
+    if (stillPending.length > 0) {
```
formaceft-93, line 659: This feels risky to me: if finalization still has pending uploads, we log the failure but continue with stopUpload() and may close the launch afterwards. That can leave TestOps with a completed/closed launch even though some results, global errors, attachments, or quality gates were never uploaded. I think we should avoid autoclosing the launch when stillPending is not empty, and probably avoid reporting a successful final status in that case.

packages/plugin-testops/test/plugin.test.ts
```diff
@@ -1583,6 +1692,61 @@ describe("testops plugin", () => {
...
+        AllureStoreMock.prototype.fixturesByTrId.mockResolvedValue([]);
+        // 1 initial + 3 retries during the normal upload, then 1 more + 3 retries during
+        // finalization: still down the whole time
+        for (let i = 0; i < 8; i += 1) {
+          TestOpsClientMock.prototype.uploadTestResults.mockRejectedValueOnce(serviceDownError);
+        }
+
+        await plugin.done({} as PluginContext, store);
+
+        // nothing throws, done() still completes, and the launch still gets stopped/closed
```
formaceft-93: This assertion seems to lock in the behavior that worries me above: done() completes and the upload still gets stopped/closed even when the deferred batch never made it through finalization. I’d rather have the test describe the safer behavior: unresolved uploads should prevent autoclose, or at least make the final state visibly incomplete/failed.

formaceft-93 (approved):

formaceft-93 (approved):

PR #907, opened 2026-09-01
resolutions ui

formaceft-93:
Added UI support for resolution categories in Awesome report: tree filter, tree markers, report/test-result tabs, and demo coverage for issue/muted/accepted cases.

Also adjusted the success-rate pie chart so muted and accepted failures are excluded from the active stats, while the full result counters/tree still show the original test statuses.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


epszaw (changes requested):
Let's move the tab before the Quality gates one and rename it just to Resolutions. It'll have the same sense but with the less polluted UI:

Image

Resolution categories accordions arrows have offset. It's possible to toggle them only when clicking the title, so the rest space is unclickable.

Image

It's not obvious what we're copying there. I think, we need to move the functionality to the resolutions space or get rid of it, because I don't know any good use case of the feature.

Image

epszaw (approved):

PR #920, opened 2026-09-02
Add Severity filter

### LINKED ISSUE #921: Filter test results by severity in the awesome report

When triaging a large report, the first question is usually "what broke that actually matters?". The severity of a test result is already visible in the report, but there is no way to narrow the result tree down to it — the quick filters cover retry, flaky, transition, tags and categories only.

Two workflows suffer from this:

- Triage: after a failed run, we want to look at the `blocker` and `critical` results first and ignore the rest until those are handled. Today that means scrolling through hundreds of results and reading each severity badge.
- Label hygiene: we cannot find the test results that carry no `severity` label at all, so nobody notices when new tests are added without one.

A severity filter in the report's quick filter bar, working like the existing transition filter:

- A dropdown listing all severity levels (`blocker`, `critical`, `normal`, `minor`, `trivial`), with multi-select — picking several levels shows the union of them.
- An additional option for test results **without** a severity label, so "everything that is not labelled" is a first-class choice rather than something you have to infer.
- The dropdown is always available, regardless of whether the current report happens to contain severity labels.
- The selection is reflected in the url, so a filtered view can be shared with the team the same way the other filters can.

### Alternatives considered

- Generating the report with `groupBy: ["severity"]`: this replaces the suite structure of the tree, needs a report regeneration, and still cannot show two severity levels together.
- Grouping by severity in the Categories tab: only covers results that match a category, so it is not a substitute for filtering the full result tree.
- Duplicating the severity as a tag so that the tags filter can be used: works, but it means maintaining the same information twice in every test.
- Searching for the severity text: the search covers labels, but it cannot express "any of these two levels" or "no severity at all".

### Additional context

_No response_

d-braun:
### Context

Fixes #921

The awesome report could be narrowed down by transition, tags and categories, but not by severity — even though the severity label is already shown on every test result. This adds a severity dropdown next to the existing quick filters.

- The dropdown is always visible and offers all five severity levels plus a separate **No severity** option, so test results that carry no `severity` label can be filtered for explicitly.
- Selecting several levels combines them with `OR`, and the selection is kept in the `severity` url param, in line with the transition filter.
- To make this work, the generator now writes the value of the `severity` label to each tree leaf, falling back to the new `noSeverityValue` (`"none"`) sentinel from `core-api` when the label is absent.
- All 21 locales already had `filters.severity` and the `severity.*` level names, so only the new `severity.none` entry had to be translated.

While covering the new filter, the same code path turned out to be broken for transitions: the transition filter rebuilt from the url put `logicalOperator` inside the field instead of on the field filter itself, so `buildFieldFilters` fell back to `AND` and chained the selected transitions with it. Since a test result has exactly one transition, selecting more than one transition matched nothing at all. That is fixed in a separate commit, with a regression test.

Bildschirmfoto 2026-09-02 um 16 14 03

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


vdvukhzhilov (changes requested):
Thank you for the contribution! 🙇 

Before we can merge this, could you please:

- Update the branch and resolve any conflicts with latest changes in `main`
- Rework the handling of tests without an assigned severity

My concern is the introduced `noSeverityValue`. By default, most test results do not have an assigned severity, so this implementation would write `severity: "none"` to almost every generated tree leaf.
Since tree data is serialized into report files, this adds unnecessary data, potentially across multiple generated trees, to represent information already expressed by the absence of the optional `severity` property.

Please keep severity absent when no severity label is assigned, and adjust the filtering implementation so that `"No severity"` can select leaves where the property is missing, without introducing or serializing a sentinel value such as `noSeverityValue`.

Thanks again for working on this! 👍

d-braun:
@vdvukhzhilov  Thanks, good catch! The sentinel is gone.

`severity` is now only written when a severity label is assigned, so leaves without
one don't carry the property at all and `noSeverityValue` is removed from `core-api`.

"No severity" now matches the *absence* of the property: `web-commons` got a generic
`NullField` filter type that builds the AQL condition `key = null` (existing AQL
comparison already treats `null` and missing alike), and the tree filter store
translates the option into it when building the predicate. `"none"` only remains as
the UI/URL option key (`?severity=none`) and never reaches the report files.

Tests updated accordingly.

vdvukhzhilov (approved):
Thanks for the update 👍 

Verified locally:

- Severity filter is available with all levels + `No severity`
- Filtering updates the tree and the severity URL param as expected
- Multi-select works as `OR`
- Leaves only get a severity field when a label is set

LGTM 🎉

PR #924, opened 2026-09-03
Fix quality gate silently passing when historyPath is set (fixes #895 and #827) (closes #861)

### LINKED ISSUE #895: Allure Quality Gate doesn't work with the config file anymore
### What happened?

I am using the allure quality-gate command to run my quality gate rules against my tests results with the following command:
`npx allure quality-gate ./allure-results --config ../../libs/testlib/testlib/allure_report/allurerc.mjs`
or
`npx allure quality-gate ./allure-results -c ../../libs/testlib/testlib/allure_report/allurerc.mjs`

i have a allure-results folder and no allure-report folder

my allurerc.mjs contains the config shown in the official documentation:
```qualityGate: {
    rules: [
      {
        // use id field to make ruleset's rules easier to identify
        id: "fast-failing ruleset",
        maxFailures: 1,
        fastFail: true,
      },
      {
        id: "non-fast-failing ruleset",
        successRate: 0.95,
      },
    ],
  },
```

it generates a allure-report folder containing one file 'test-results.json'. This file only contains this:

```{
    "byId": {
        "mytestid1": {
            "id": "mytestid1",
            "name": "my test name",
            "duration": 19562,
            "status": "failed"
        },
        "mytestid2": {
            "id": "mytestid2",
            "name": "my test name",
            "duration": 19321,
            "status": "failed"
        },
    }
}
```

i have 2 tests and both are KO, so both rules should fail but nothing is printed to the console.

However if i run the quality gate with the inline rule arguments, i get this:
```
npx allure quality-gate ./allure-results --max-failures 1
Quality Gate failed with following issues:

 ⨯ The number of failed tests 2 exceeds the allowed threshold value 1    maxFailures

1 quality gate rules have been failed.
```

if i run `npx allure quality-gate ./allure-results` i get an error message "Quality gate is not configured!
Add qualityGate to the config or consult help to know, how to use the command with command-line arguments"

and if i run `npx allure generate ./allure-results --config ../../libs/testlib/testlib/allure_report/allurerc.mjs` to generate a allure-report folder and the run `npx allure quality-gate ./allure-results --config ../../libs/testlib/testlib/allure_report/allurerc.mjs`, nothing is printed to the console and the same "test-results.json" file i mentioned is created inside my existing allure-report folder

So my guess is that the config file is recognized, but the rules defined in the config file aren't used

For reference, this was working properly in allure 3.14.0

### How can we reproduce it?

Using the config from the official documentation in my allurerc.mjs config file:
```

export default defineConfig({
  qualityGate: {
    rules: [
      {
        // use id field to make ruleset's rules easier to identify
        id: "fast-failing ruleset",
        maxFailures: 5,
        fastFail: true,
      },
      {
        id: "non-fast-failing ruleset",
        successRate: 0.95,
      },
    ],
  },
});
```

Run 2 tests that are KO (i am using playwright typescript with the allure-playwright reporter) > this should generate the allure-results folder

Run npx allure quality-gate ./allure-results -c <path to the allurerc.mjs file>
The command runs and shows nothing in the terminal
a allure-report folder was created and contains 1 single file test-results.json

### What did you expect?

The command runs and shows the results of the quality gate defined in the config file against the tests results from the allure-results folder

### Environment

- Allure 3 version: 3.16.0
- Node.js version: v22.11.0
- OS: Windows
- @playwright/test": "1.59.1",

### Additional context

I need the console print to work with a config file because i want to declare custom rules, and i cant run my tests directly using allure run because i have some data enrichment that needs to run between my test run and my allure report generation

d-braun:
### Context

Fixes #895 and #827 (closes #861).

When `historyPath` is configured, `allure quality-gate` exits with code **0** and prints nothing, no matter how many rules the results violate. The only trace it leaves is an `allure-report/` directory containing a single `test-results.json` and no `quality-gate.json`, exactly as reported in the issue. `allure generate` and `allure run` hit the same defect; there it goes mostly unnoticed because the report has already been written by the time the process stalls.

**Root cause.** `AllureLocalHistory` opens streams on a `FileHandle` with `autoClose: false` and then awaits `FileHandle.close()` in a `finally` block without ever destroying them. Such streams are never destroyed on their own, and the handle stays referenced until they emit `"close"` — so `FileHandle.close()` waits for a reference that is never released and never settles. As a consequence `AllureReport.done()` never returns, the event loop drains, and Node exits with code 0 before the quality gate is ever validated.

This is why the issue looks like the config file is ignored: the reporter's config sets `historyPath`, while the `--max-failures` run they compared against was made *without* `--config` and therefore without history. Rule parsing was never the problem.

Note that running the CLI through `yarn allure …` hides the defect — Yarn PnP preloads `.pnp.cjs`, which shifts the timing enough for `close()` to settle. Under plain `node` / `npx`, which is what users run, it deadlocks. Vitest masks it as well. The numbers below come from a clean `npm i allure@3.16.0` with the built `history.js` swapped in and run under plain `node`:

| Scenario | Before | After |
| --- | --- | --- |
| `quality-gate` with `historyPath` | exit 0, no output, no `quality-gate.json` | exit 1, both rule violations reported |
| 4 consecutive runs, also with `historyLimit` | deadlock | exit 1, history appended and trimmed |
| `generate` with `historyPath` | exit 0, but `done()` never returned | exit 0, history grows as expected |

**Changes.**

- `packages/core/src/history.ts` — destroy every stream opened on the history file handle before closing it, in the order the streams were opened: destroying a reader that shares its handle with a still open writer never completes, whereas destroying the writer closes the reader along with it. Closing the handle tolerates `EBADF`, because a destroyed stream may already have closed the descriptor depending on the runtime.
- `packages/cli/src/commands/qualityGate.ts` — safety net: a quality gate must never pass silently. If the process is about to exit before the validation has finished, the command now reports it and exits with code 1 instead of leaving the caller with a successful exit code. Checked against the *unfixed* deadlock, it turns the silent exit 0 into a loud failure.

**Tests.**

- `packages/core/test/history.streams.test.ts` (new) — asserts that every stream opened on the history file handle is closed, for the new-file, existing-file and read paths. All three fail without the fix.
- `packages/cli/test/commands/qualityGate.test.ts` — two tests for the new safety net: it fires on an interrupted run and stays silent on a completed one.

`yarn format:check`, oxlint and the `@allurereport/core` (528) and `allure` (217) suites pass.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/core/src/history.ts
```diff
@@ -98,6 +99,39 @@ export const createHistory = (
...
+ */
+const closeStream = async (stream: Readable | Writable): Promise<void> => {
+  if (stream.closed) {
+    return;
+  }
+
+  await new Promise<void>((resolve) => {
+    stream.once("close", () => resolve());
+    // releasing the resources must not fail the operation the stream has already completed
+    stream.once("error", () => resolve());
```
formaceft-93: Could we keep this promise settled by `close` only? The guarantee we need here is that the FileHandle-backed stream has actually released its descriptor reference before `closeFileHandle()` runs. If `error` fires before `close`, this can theoretically move on too early. We can still attach an `error` listener to swallow cleanup errors, but let the promise resolve on `close`.

packages/cli/test/commands/qualityGate.test.ts
```diff
@@ -372,4 +372,84 @@ describe("quality-gate command", () => {
     expect(AllureReportMock.prototype.readDirectory).toHaveBeenCalledWith("./foo/");
     expect(AllureReportMock.prototype.readDirectory).toHaveBeenCalledWith("./bar/");
   });
+
+  describe("interrupted validation", () => {
+    const captureBeforeExitListeners = () => {
+      const listeners: (() => void)[] = [];
+
+      vi.spyOn(process, "once").mockImplementation(((event: string, listener: () => void) => {
```
formaceft-93: Can we restore this spy after each test? `vi.clearAllMocks()` clears call history but does not restore the mocked `process.once` implementation, so later tests in the same worker would keep using this captured-listener mock. It happens to be safe while these tests are last, but restoring it would make the tests self-contained.

formaceft-93 (changes requested):

formaceft-93:
@d-braun Thank you for your contribution, I left two small comments, and could you please also make a rebase on current `main` because #958 also touched local history handling.

d-braun:
Thanks @formaceft-93!
Both addressed and rebased onto current main, tests green.

epszaw:
Hello, @d-braun! Some tests windows failed. Please, resolve the problems before we continue

d-braun:
Thanks @epszaw! Fixed. :)

formaceft-93:
@d-braun Sorry to bother you again, but can you resolve conflicts one more time, please :)

PR #928, opened 2026-09-03
Show passed quality gate rules in the report (fixes #806)

### LINKED ISSUE #806: Quality Gate - Show Passed Rules

Currently, when a test run passes all Quality Gate rules, the Quality Gate tab displays "No quality gate results available". This can be misleading for report viewers, as it may give the impression that no Quality Gate rules have been configured for the project.

Additionally, when multiple Quality Gate rules are configured and only one fails, only the failing rule is displayed. This can make it appear as though only that single rule exists, while the successfully validated rules remain invisible.

From a reporting perspective, this makes it difficult for users reviewing the report to understand the full Quality Gate configuration and the overall quality assessment.

A successful Quality Gate check is still a meaningful result and should be visible.

It would be valuable if the Quality Gate tab displayed all configured rules, including both passed and failed rules. This would provide a complete overview of the quality criteria that were evaluated and make the report more transparent.

As a QA engineer, I use the Allure report to communicate automated test results to stakeholders and decision-makers. I am not the person making the quality decisions, so the report should clearly show all evaluated criteria to help others understand the outcome.

Maybe something like this could help?
Image

### Alternatives considered

I looked into the generated report files and noticed that after the report is generated, a quality-gate.json file is created at allure-report/widgets/quality-gate.json

When a Quality Gate rule fails, the file contains an entry with a structure similar to:
```
{
    "default": [
        {
            "success": false,
            "actual": XXX,
            "expected": XX,
            "rule": "XXX",
            "message": "XXX"
        }
    ]
}
```

A workaround i looked into would be to post-process this file after report generation, append successful rule results manually with "success": true, and apply custom CSS changes to make them visible in the UI.

However, this approach feels fragile and would require maintaining custom logic outside of Allure. A native Quality Gate display that includes both passed and failed rules would provide a more reliable and consistent solution for users.

### Additional context

Thank you for taking the time to look into this suggestion.

d-braun:
### Context

Fixes #806.

The Quality Gate tab only ever showed failed rules. A run that passed every rule rendered
"No quality gate results available", which reads as if no quality gate was configured at all, and a
run with one failing rule out of many hid the rules that were validated successfully. Report readers
had no way to see which quality criteria were actually evaluated.

`QualityGate.validate()` dropped passed rules with an early `continue`, so they never reached the
store and never made it into `widgets/quality-gate.json`. This changes the validation to report every
evaluated rule and makes the report render passed and failed rules side by side.

**Core (`@allurereport/core`, `@allurereport/plugin-api`)**

- `QualityGate.validate()` now returns every evaluated rule, each carrying its `success` flag.
  `fastFail` still triggers on failures only.
- New `filterFailedQualityGateResults()` helper. `stringifyQualityGateResults()` and
  `convertQualityGateResultsToTestErrors()` filter internally, so terminal output and test errors keep
  reporting failures only.
- `QualityGateRule` gained an optional `successMessage`. All built-in rules provide one; custom rules
  without it fall back to a generic message.

**CLI (`allure`)**

- The exit code of `allure quality-gate` and `allure run` is now derived from the number of *failed*
  rules instead of the length of the result list. Passed rules are sent to the report but are not
  printed to the terminal.

**Report (`@allurereport/web-awesome`)**

- Passed rules render with a green check icon and a green "Success" block; failed rules are unchanged.
- `TrError` accepts an optional `title` so the block header can read "Success" instead of "Error".
- The tab counter keeps its previous meaning and counts failed rules only, so it stays red only when
  something actually failed, while the tab itself lists every evaluated rule.
- Added the `ui.success` key to all 21 locales.

No changes to the shape of `widgets/quality-gate.json` beyond the additional entries: passed rules
appear as regular entries with `"success": true`, so existing consumers keep working.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

Bildschirmfoto 2026-09-03 um 13 39 05
Bildschirmfoto 2026-09-03 um 13 39 13


auranemartin:
This looks incredible, thank you for taking the time to work on my feature request 🫶

I would just suggest that the quality gates count in the tab name doesnt only increase when a quality gate rule is failed.
On your screenshot, it shows that when the 3 rules are successful, the count in the tab is 0 which can be kinda confusing when looking at it from a different tab since it can look like there are no quality gates rule configured
Maybe just always have it show the count of quality gates rules, but have it in red when at least one rule is failed?

d-braun:
Alright! :)
1. The tab counter now shows **all** quality gate rules and turns
red as soon as one fails (green when all pass, neutral when none evaluated).
2. Failed rules are also listed first, keeping the original order within each group.
Bildschirmfoto 2026-09-08 um 18 36 55
Bildschirmfoto 2026-09-08 um 18 37 04

epszaw (approved):

epszaw:
@d-braun thank you for a wonderful improvement!

PR #931, opened 2026-09-04
feat(core): write local artifacts manifest

formaceft-93:
- Add a local-only `artifacts.json` manifest for report generation artifacts.
- Record dump restore outcomes, including missing, failed, restored, and nested expanded dumps.
- Record runtime-config global attachment patterns and matched in-cwd files.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/core/src/report.ts
+++ b/packages/core/src/report.ts
@@ -4,7 +4,7 @@ import { once } from "node:events";
 import { createReadStream, createWriteStream, existsSync, readFileSync, type ReadStream } from "node:fs";
 import { lstat, mkdtemp, readdir, realpath, rename, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
-import { basename, dirname, join, relative, resolve, sep } from "node:path";
+import { basename, dirname, join, resolve } from "node:path";
 import { promisify } from "node:util";
 
 import { detect } from "@allurereport/ci";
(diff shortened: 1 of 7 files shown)


packages/core/src/report.ts
```diff
@@ -74,7 +74,38 @@ const DEFAULT_READ_CONCURRENCY = 64;
...
-const ROOT_INTEGRATION_FILENAMES = new Set([TEST_RESULTS_REGISTRY_FILENAME, QUALITY_GATE_RESULTS_FILENAME]);
+const ARTIFACTS_MANIFEST_FILENAME = "artifacts.json";
+const ROOT_INTEGRATION_FILENAMES = new Set([
+  TEST_RESULTS_REGISTRY_FILENAME,
+  QUALITY_GATE_RESULTS_FILENAME,
+  ARTIFACTS_MANIFEST_FILENAME,
+]);
+const OUTSIDE_CWD_ARTIFACT_PREFIX = "<outside-cwd>";
+
+type ReportArtifactsDump = {
```
epszaw: Let's move the constants, types, and functions that don't work with the report core directly to the `packages/core/src/utils/artifacts.ts` file to make the report as clean as possible.

epszaw (changes requested):
The goal is to list the artifacts actually used during report generation. The current implementation adds more bookkeeping than that needs: dump statuses, errors, and configured glob patterns turn the manifest into a restoration log. A single list of `{ name, path }` entries should be enough, containing successfully consumed dumps and attached files. Unmatched patterns should contribute nothing, and failures can remain in the existing logs.

The commands already resolve dump patterns before calling `restoreState()`, so recording belongs in `AllureReport`, where we know which inputs were consumed. We should also deduplicate dump paths before restoration. For archives containing nested dumps, recording the original archive once should be sufficient; temporary extracted files don’t need their own entries.

Path handling can be simpler too. Use paths relative to the generation `cwd`, including `../` where needed, instead of `<outside-cwd>`. That preserves the source location and lets us reconstruct it with `resolve(cwd, path)`. Dump matching, global attachment matching, and manifest paths should all use that same working directory.

formaceft-93:
I added cwd to FullConfig as runtime context from readConfig(cwd, ...), so dump matching, global attachment matching, and manifest path normalization use the same working directory. This is intentionally not a user config field in .allurerc.

epszaw (approved):

epszaw:
Everything works as expected. Dump files and global attachments are included to the artifacts manifest:

```
$ cat allure-report/artifacts.json

[{"name":"foo.zip.zip","path":"foo.zip.zip"},{"name":"bar.txt","path":"bar.txt"}]%
```

PR #939, opened 2026-09-07
Disable Publish PR reports for fork PR and pack TestOps CI overrides

vdvukhzhilov:
### Context

- Skip PR Allure report publishing on PRs from forks as this job requires secrets and permissions not available in fork PRs context.
- Enable `gitFlow` in `allurerc.mjs`
- Extend GitHub CI detection for `workflow_run` PR context, and expand TestOps overrides (`ALLURE_CI_*` for git/PR/repo/commit, `ALLURE_JOB_*` for job identity) so the publisher can attribute launches to the original Build project run
- Allow `/update-bundle-baseline` on forks with measure + artifact/PR-comment handoff instead of pushing to the fork

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/.github/workflows/update-allure-bundle-baseline.yml
+++ b/.github/workflows/update-allure-bundle-baseline.yml
@@ -19,8 +19,6 @@ on:
 
 permissions:
   contents: read
-  issues: write
-  pull-requests: write
 
 concurrency:
   group: update-bundle-baseline-${{ inputs.pr_number }}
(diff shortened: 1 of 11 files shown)


.github/workflows/update-allure-bundle-baseline.yml
formaceft-93: This artifact is uploaded for fork PRs before the baseline is validated. Since measure runs PR-controlled code, it can produce an invalid or modified JSON file, which is then handed off to the user for manual replacement. Could we validate and canonicalize the baseline before uploading it, so both fork handoff and same-repository commit use the same validated artifact?

formaceft-93 (approved):

PR #955, opened 2026-09-09
Move history ref creation to report generation step

andrcuns:
### Context

Removes full history url creation at runtime in the browser and moves it to report generation phase. This allows to separate url behaviors for local history and allure history service which may behave differently. (report storage service always stores report at path that includes plugin id regardless if multiple or just single report plugin is configured which won't be the case for manual hosting where path will be flattened in case of single reporter plugin).

This allows to correctly set history link at generation time and not couple runtime behavior to report storage service.

#### Checklist

- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


.tool-versions
```diff
@@ -0,0 +1 @@
+node 24.14
```
andrcuns, line 1: I added this file because our tests don't actually work with node 26 which was my default environment and that wasted a lot of agent tokens it trying to debug these issues.

`.tool-versions` file is read by various managers like [mise](https://mise.jdx.dev/) and [asdf](https://asdf-vm.com/) and should help avoid such issue.

packages/plugin-awesome/src/history.ts
```diff
@@ -0,0 +1,11 @@
+export const buildServiceHistoryUrl = (historyUrl: string, pluginId: string, historicalResultId: string): string => {
+  if (!historyUrl) {
+    return "";
+  }
+
+  const { origin, pathname } = new URL(historyUrl);
```
formaceft-93, line 6: Could we make this non-throwing? `HistoryTestResult.url` comes from persisted history data and is not validated as an absolute HTTP URL before this point. With `new URL(historyUrl)` here, one malformed/custom history entry can fail the whole Awesome report generation. I think we should parse defensively, probably accept only http/https report roots, and return an empty URL for invalid values so the item stays visible but unlinkable.

packages/plugin-awesome/src/history.ts
```diff
@@ -0,0 +1,11 @@
...
+  const navigateUrl = new URL([pathname, pluginId].join("/"), origin);
```
formaceft-93, line 7: Can we avoid preserving the double slash here? A stored report root ending with `/` is a normal hosted-directory URL, but `[pathname, pluginId].join("/")` turns it into `/run-123//awesome`. Some static hosts/CDNs treat that as a different path, so the history link may 404. I'd rather trim trailing slashes from the base path before appending `pluginId` and update this test to expect `/run-123/awesome`.

andrcuns:
@formaceft-93 Thanks for feedback, it was pretty fully generated by LLM and on first review it made sense but then I combined it with the second part and it stopped making sense 😅 

I will address the logical feedback, close this mr and instead of the stack, open as single mr.

formaceft-93:
> @formaceft-93 Thanks for feedback, it was pretty fully generated by LLM and on first review it made sense but then I combined it with the second part and it stopped making sense 😅
> 
> I will address the logical feedback, close this mr and instead of the stack, open as single mr.

I'm also doing this one https://github.com/allure-framework/allure3/pull/831
It overlaps at the history-link composition boundary. 
But here for Allure service storage, appending the plugin id during generation makes sense because storage keeps plugin reports under /<pluginId>. For executor/Jenkins/manual history, especially the default single Awesome report, the generated output is flattened, so the browser must not append /awesome blindly. 

We should keep these two cases separated and avoid composing the same history href twice.

andrcuns:
@formaceft-93  "so the browser must not append /awesome blindly.", yes, that was the main issue that prevented local history from working correctly. I completely separated the concerns in my latest branch given we actually already have the separate Local history and Service history classes. It's just I eventually noticed that the code llm generated decided not to take advantage of that and still generated a sort of a common solution.

PR #956, opened 2026-09-09
Add option to set history url base for local history provider

andrcuns:
### Context

This adds a new historyBaseUrl configuration option which allows to fully control history url creation when local history provider is used (user fully controls how reports are hosted and thus can provide the base url for correct history url creation)
#### Checklist

- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/packages/cli/src/commands/allure2.ts
+++ b/packages/cli/src/commands/allure2.ts
@@ -65,6 +65,10 @@ export class Allure2Command extends Command {
     description: "The path to history file",
   });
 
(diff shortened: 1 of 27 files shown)


andrcuns:
Moved to draft as I need to double check a few things that doesn't seem right

PR #958, opened 2026-09-10
Add support for local history via historyBaseUrl parameter

andrcuns:
### Context

This PR adds support for properly functioning local history with awesome reporter via the following changes:

* Remove dynamic history link building in awesome report and instead build final url during report generation
* Separate how history url is built for local and service histories due to difference in hosting logic (report with local history has no predetermined hosting strategy, we only assume that the final report path will match folder structure of output, service has predetermined hosting structure and always hosts at path "$pluginId/" path). History links for local history are built only when base url is set. The history url is left blank otherwise as before.
* Add additional config parameter historyUrlBase which is only used when historyPath is present and local history is used. The base path allows to define the base of the report it will be published at.
* Move recent bug fix of handling missing trailing slashes from client to report generation

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/core/src/history.ts
```diff
@@ -108,6 +141,22 @@ export class AllureLocalHistory implements AllureHistory {
...
   ) {}
 
+  resolveTestResultUrl(historyUrl: string, pluginId: string, historicalResultId: string): string {
+    if (!historyUrl) {
+      return "";
+    }
+
+    const url = new URL(historyUrl);
+
+    if (url.pathname.endsWith("/")) {
```
andrcuns: Local history always normalizes the url, so it will be either the base path ending with `/` in case of multireport config or `.html` in case of single report configuration

packages/cli/src/commands/history.ts
```diff
@@ -42,6 +42,10 @@ export class HistoryCommand extends Command {
     description: "Limits the number of history entries to keep (default: unlimited)",
   });
 
+  historyUrlBase = Option.String("--history-url-base", {
```
epszaw: Maybe rename it to `historyBaseUrl`?

epszaw (approved):

PR #961, opened 2026-09-10
feat(cli): print quality gate results as json

### LINKED ISSUE #807: Quality Gate - Option to print the quality-gate.json content in the console

Currently when running `npx allure quality-gate`, the console only prints the error message, which is fine when you have the full allure report generated but doesn't really help when exporting results as a csv only for example.

It would be great if we had a flag like --print-results to print the quality gate results the way it is formatted in the allure-report\widgets\quality-gate.json file when using `npx allure run`
having it printed to the console would then allow users to process it using a script or command lines and append it to another file like their csv export for example

### Alternatives considered

I have a script that runs after my allure report was generated based on the exported data, but that means im not using the quality-gate feature at all since i have to compute the results with my script to be able to include it in my export

### Additional context

_No response_

formaceft-93:
Adds a --print-results option to allure quality-gate so scripts can read the same JSON-shaped validation results that are written to quality-gate.json. The default terminal output and exit behavior stay unchanged; with the flag, stdout always receives a JSON array, including [] when validation passes.

Closes #807.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


epszaw (changes requested):
It's better to print quality gate results as a formatted message, like we do this for test results, but not just as raw json data.

Let's also add an option to turn off the results logging in the `plugin-log`.

packages/cli/src/commands/qualityGate.ts
```diff
@@ -178,6 +182,10 @@ export class QualityGateCommand extends Command {
       environment: resolvedEnvironment?.id,
     });
 
+    if (this.printResults) {
```
epszaw: Let's remove the option because we already added an option to print quality gate results using the **log** plugin.

If users need to check `.json` result file, they always can inspect the `<report_output_dir>/quality-gate.json` file (e.g. `cat ./allure-report/quality-gate.json`). The option in the log plugin should solve the initial problem well enough.

epszaw (changes requested):

epszaw (approved):

PR #971, opened 2026-09-14
fix(summary): use wall clock run duration on report widgets

### LINKED ISSUE #969: Incorrect run duration on multi-report widgets
### What happened?

When generating a report with several plugins, a report selector appears with one widget per plugin
These widget are displaying some information about the test run, including its duration

However the duration on the widget doesn't match the duration of the pipeline nor the duration displayed on the specific plugin reports:

Image

Image

I am running my test using several workers in playwright, it looks like the widget duration is the sum of all tests durations, whereas the report duration is the actual execution duration

### How can we reproduce it?

- Have a config with different plugins (in my case, i have 2 awesome, 1 allure classic, 1 allure2)
- Run playwright tests **with several workers**  (in my case, 5)
- Generate the report

### What did you expect?

The duration displayed on the widgets should match the duration displayed on the actual reports

### Environment

- Allure 3 version: 3.17.0
- Node.js version: 22.11.0
- Playwright version: 1.63.0

### Additional context

_No response_

todti:
Closes #969

Report selector widgets showed the sum of individual test durations, so with parallel workers the number was several times larger than the real run time and did not match the duration shown inside the reports themselves.

`createPluginSummary` now derives the duration from the span between the earliest start and the latest stop of the test results — the same way `getRunSummary` does for the awesome report. When results carry no start/stop bounds (e.g. junit.xml input), it still falls back to the sum of durations.


packages/plugin-api/src/utils/summary.ts
```diff
@@ -27,6 +27,24 @@ export const convertToSummaryCheckResult = (check: AllureCheckResult): SummaryCh
   status: check.status,
 });
 
+export const calculateRunDuration = (testResults: Pick<TestResult, "start" | "stop" | "duration">[]): number => {
+  let start = Infinity;
+  let stop = -Infinity;
+
+  for (const { start: trStart, stop: trStop } of testResults) {
+    if (typeof trStart === "number" && typeof trStop === "number") {
```
formaceft-93: Small nit: should we use `Number.isFinite` for `trStart`/`trStop` here, same as `getRunSummary`, to keep invalid timings like `NaN`/`Infinity` out?

formaceft-93 (approved):

packages/plugin-api/src/utils/summary.ts
todti: Good catch — switched to `typeof x === "number" && Number.isFinite(x)` for both bounds, same as `getRunSummary`, and added a test with `NaN`/`Infinity` timings. Fixed in 85e6957.

epszaw (approved):

PR #972, opened 2026-09-14
Make breadcrumb path items clickable (fixes #501)

### LINKED ISSUE #501: [Awesome Report] Breadcrumb path items not clickable - navigation broken
### Description
In Allure 3 Awesome Report, breadcrumb navigation items (showing test suite hierarchy) are rendered as plain text without click handlers. Users cannot navigate back through the test hierarchy by clicking on breadcrumb items.

### Expected Behavior
Clicking on breadcrumb path items (e.g., "com.example.tests > LoginPageTest > shouldLogin") should navigate to that level in the test hierarchy.

### Actual Behavior
Breadcrumb items are static text. Only the home icon (🏠) is clickable for navigation back to root.

### Investigation
Examining minified source code of Allure 3.1.0:
- Breadcrumb path items use `Cc` component (plain text)
- No `onClick` handlers attached
- Only home button has `onClick:()=>t4()` handler

### Steps to Reproduce
1. Open any Allure 3 Awesome Report
2. Navigate to a test case (not at root level)
3. Try to click on breadcrumb items to go back
4. Observe: items are not clickable, only home icon works

### Environment
- Allure version: 3.1.0
- Report type: Awesome Report
- Browser: Any

### Related Issues
- #56 - Navigation improvements
- #228 - Breadcrumb issues
- #306 - Test tree navigation

### Proposed Solution
Add click handlers to breadcrumb path items to enable navigation to any level in the hierarchy.

todti:
### Context

Fixes #501.

Breadcrumb segments were plain text — only the home icon navigated anywhere.

They were also built from `parentSuite`/`suite`/`subSuite` labels regardless of how
the report tree is actually grouped. With a custom `groupBy`, or with a `titlePath`
tree, the crumbs named levels that do not exist in the tree at all, which is why
wiring clicks to them had no obvious destination.

The path now comes from the tree the user is looking at, so every segment maps to a
real node — the results tree for a test result opened from the tree, the categories
tree for one opened from categories. Clicking a segment expands its ancestors,
focuses it and scrolls it into view. In split mode the test result stays open;
in full-page mode the click returns to the tree. A test result that is missing from
the tree keeps rendering the label breadcrumbs as plain text, so nothing regresses.

Paths longer than four segments collapse the middle into a menu that opens on hover
or click, keeping the root and the two levels above the test readable. Segments no
longer shrink proportionally — on deep paths that used to chop every one of them
mid-word.

One incidental fix came out of this: category nodes kept their expanded state in
local component state seeded from `collapsedTrees` at mount, so an already mounted
tree ignored external changes and needed a forced remount after subtree toggles.
It is derived from the signal now, and the remount is gone.

#### Checklist

- [ ] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/web-awesome/src/stores/tree.ts
```diff
@@ -190,3 +191,55 @@ export const filteredTree = computed(() => {
...
+ * Path of the test result through the report tree, as it's actually grouped,
+ * so every segment points to an existing tree node.
+ */
+export const getTreeBreadcrumbs = (testResultId?: string): TreeBreadcrumb[] => {
+  if (!testResultId) {
+    return [];
+  }
+
+  for (const tree of Object.values(treeStore.value.data ?? {})) {
+    const path = findBreadcrumbs(tree.root, tree.groupsById, testResultId, []);
```
formaceft-93: Should this use filteredTree instead? Breadcrumbs are built from the full treeStore, but revealTreeNode looks in the filtered tree. With search/filter enabled, we can show a clickable breadcrumb that is not actually visible in the tree.

formaceft-93 (approved):

packages/web-awesome/src/stores/tree.ts
todti: Good catch. `getTreeBreadcrumbs` now walks `filteredTree`, exactly the tree `revealTreeNode` walks. With search or filters on, a test result that is not in the tree falls back to the plain-text label breadcrumbs instead of offering a dead link.

epszaw (approved):

PR #975, opened 2026-09-15
Implement "gitlab generate" command

andrcuns:
### Context

Implements `gitlab generate` command which is responsible for creating allure reports with functional history published using gitlab artifacts with option to create a report summary comment.

The command is designed to run within a gitlab CI job and do the following:

* Detect relevant previous pipeline and fetch history.jsonl from the job with the same name before report generation
* Generate allure report respecting existing config file and cli parameters
* Print test result summary and link to the report as merge request comment when executed in context of one. Print report log in job log output when not running within mr
* Place both updated history and generated report where gitlab ci job can upload it as artifacts

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


andrcuns:
@formaceft-93 @epszaw Would you mind taking a look? It's a bit on the larger side because I was experimenting with agent setup and forgot to instruct it to pause for human reviews so it implemented the whole thing.

The code seems generally ok, there is a bit too much of all kinds of edge case handling, but I left it in. While it's being reviewed, I want to try and build the cli from branch and test it E2E on real GitLab project, so maybe some issues might still come up.

andrcuns:
@formaceft-93 @epszaw FYI, I pushed some code that simplified the generate command and completely removed the logic for summary comment which will be added as a shared component in https://github.com/allure-framework/allure3/pull/983

packages/cli/src/commands/gitlab/generate.ts
```diff
@@ -0,0 +1,143 @@
...
+
+const reportBaseUrl = (historyBaseUrl: string | undefined, output: string): string => {
+  const gitlab = detect();
+  const isGitlabCiDescriptor = (ci: CiDescriptor): ci is GitlabCiDescriptor => ci.type === CiType.Gitlab;
+
+  if (!isGitlabCiDescriptor(gitlab)) {
+    throw new Error("GitLab CI environment was not detected");
+  }
+
+  return historyBaseUrl || `${gitlab.jobArtifactsUrlBase}/${output}`;
```
formaceft-93, line 52: `historyBaseUrl` is built by appending `output` to the GitLab artifacts URL. That works for repo-relative output paths, but if `--output` is absolute, the generated MR/history link becomes something like `.../artifacts//builds/group/project/report/index.html`, which GitLab artifacts won’t serve.

I think we should normalize `output` to a path relative to `CI_PROJECT_DIR` before building the default URL, and either reject outputs outside the project or require `--history-base-url` for them.

andrcuns: Makes sense. It wouldn't happen in the current gitlab ci component because output is actually hardcoded because there is no real case to change it, but functionally it is indeed more correct to make sure 👍 

I found a few more minor issues after doing e2e test, I will push a fix with my next commit

andrcuns: This should now be addressed

andrcuns:
@epszaw @formaceft-93 Sorry for a slightly premature review ask. I pushed a few more commits with some slop removal and minor logging fixes and addressed the issue from review. I think it's ready to be reviewed now.

Also ran the code from this branch on a real GitLab merge request, it works with a fully functional history - https://gitlab.com/dependabot-gitlab/dependabot/-/merge_requests/5208#note_3843609767

packages/cli/src/commands/commons/summary.ts
```diff
@@ -0,0 +1,28 @@
+import type { Statistic } from "@allurereport/core-api";
+import type { AllureStore } from "@allurereport/plugin-api";
+
+export const collectReportSummary = async (
```
epszaw, line 4: We already have the `createPluginSummary` function in the `plugin-api` package that implements the same logic.

We can modify the function and reuse it instead of authoring a new one.

packages/ci/src/helpers/gitlab/client.ts
```diff
@@ -0,0 +1,295 @@
...
+const bodyCancel = (response: Response) => {
+  void response.body?.cancel().catch(() => undefined);
+};
+
+const requestWithTimeout = async <T>(
+  url: URL,
+  init: RequestInit,
+  consume: (response: Response) => Promise<T>,
+): Promise<T> => {
+  const response = await fetch(url, {
```
epszaw: Native `fetch` is available starting from `node@21`, so I recommend to use `axios` instead (we already ship the package during the installation).

packages/cli/src/commands/gitlab/generate.ts
```diff
@@ -0,0 +1,177 @@
...
+        await operation();
+      } catch (error) {
+        err(`${errPrefix}: ${error instanceof Error ? error.message : String(error)}`, this.context);
+      }
+    };
+
+    log("Generating allure report", this.context);
+    log("  fetching previous run history", this.context);
+    await runGitlabOperation("  history fetch failed", () =>
+      restoreGitlabHistory({
```
epszaw: Don't we want to support fetching history from the Allure Service if it's present in the config and the history is available?

packages/cli/src/commands/gitlab/generate.ts
```diff
@@ -0,0 +1,177 @@
...
+import { cwd as processCwd } from "node:process";
+
+import { restoreGitlabHistory, upsertGitlabJobNote, detect, GitlabCiDescriptor } from "@allurereport/ci";
+import { readConfig } from "@allurereport/core";
+import { CiDescriptor, CiType } from "@allurereport/core-api";
+import { BaseContext, Command, Option } from "clipanion";
+
+import { generate } from "../commons/generate.js";
+
+const defaultHistoryLimit = 100;
```
epszaw: Let's use upper snake case for constant definitions

packages/cli/src/commands/gitlab/generate.ts
```diff
@@ -0,0 +1,177 @@
...
+      throw new Error("Absolute output path must be within CI_PROJECT_DIR");
+    }
+
+    output = relative(projectDirectory, absoluteOutput).split(sep).join("/");
+  }
+
+  return historyBaseUrl || `${gitlab.jobArtifactsUrlBase}/${output}`;
+};
+
+const log = (msg: string, context: BaseContext) => {
```
epszaw: Plain `console` methods won't work in the case?

epszaw (changes requested):

packages/cli/src/commands/commons/summary.ts
andrcuns: Yes, this was added before merging the shared CI component, this file is completely removed in the other stacked pr: Yes, this was added before the shared summary component, removed in the other stacked pr (checks failed due to unrelated reasons so couldn't yet merge it): https://github.com/allure-framework/allure3/pull/997/changes#diff-ba79f9dd9d1bbb8b41c01fb59e59b4a28cdc2376d6daac936aa3e23b173a0736

packages/cli/src/commands/gitlab/generate.ts
andrcuns: They probably do. This construct was chosen by llm and I didn't really question it to be honest. My guess is that the assumption for this is that if context implements some special output setup, we wouldn't need touch this command to reuse it. But it is probably overoptimisation ahead of time

packages/cli/src/commands/gitlab/generate.ts
andrcuns: I thought that with initial scope we would focus on local history and report publishing via artifacts and add the service support as a follow up to somewhat limit the amount of changes, but we can widen the scope and support it right away.

packages/ci/src/helpers/gitlab/client.ts
andrcuns: Updated to axios and also removed whole bunch of manual http layer handling because axios has built in utilities, like stripping token on redirect to aws etc.

packages/cli/src/commands/gitlab/generate.ts
andrcuns: Replaced with basic console calls

packages/cli/src/commands/gitlab/generate.ts
andrcuns: Updated behavior to check if service token is set and skip local history retrieval if it is. The urls will be correctly handles at report generation stage so the only guard we need is to skip actual history.jsonl fetching

epszaw (approved):

Issue #224, opened 2025-06-24
Indicate tests that changed their status in a retry


delatrie:
### The context

Currently, the `retried` indicator is shown next to a test if it has at least one retry, regardless of the outcomes.

In Allure Awesome, such an indicator is shown only if the test has different outcomes across its attempts.

Both implementations have their own advantages:

  - The current implementation allows assessing the impact the retrying policy has on the test base.
  - The Allure 2 implementation allows assessing the efficiency of the retrying policy in terms of the number of instabilities it detects. It also keeps flaky tests on the radar better (we can distinguish between tests that passed on the first try and those that passed after a retry).

On the other hand, the current implementation has a disadvantage: it makes the report noisier in case retries are unconditionally enabled for all tests. Some users employ such a strategy to define flaky tests ([an example](https://github.com/allure-framework/allure3/issues/160#issuecomment-2996565479).

### The proposal

1. We should highlight tests with different statuses across their attempts.
2. We should show the total number of such tests.

### Open questions

We may either rework the current `retried` indication and counter or add a new one.

I personally prefer the latter as it will allow a better understanding of the efficiency of the retrying policy. For example, we'll be able to further asses the efficiency by comparing the number of flipped tests with the total number of retried tests. However, the downside is that it may make the report noisier and more difficult to read.

PR #978, opened 2026-09-15
feat(awesome): indicate retry status changes separately

formaceft-93:
## Summary

Adds a separate `retriesStatusChange` signal for tests whose visible/latest attempt has a different significant status than at least one retry attempt.

This keeps the existing retry indicators and retry totals for measuring retry-policy coverage, while adding a separate way to spot tests that actually changed outcome across attempts.

## Semantics

`retriesStatusChange` is true when:

- the latest visible attempt status is one of `passed`, `failed`, or `broken`;
- at least one previous retry attempt also has a significant status;
- that previous significant status differs from the latest visible attempt status.

`skipped` and `unknown` do not create a status change and are not used as comparison bases.

Issue: https://github.com/allure-framework/allure3/issues/224

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


delatrie (commented):
#### Summary title

I have two issues with the summary title:

Image

1. The text is quite long.
2. It reads as we're counting status transitions across retries rather than tests with at least one such transition.

I don't have a perfect text in mind. Something like "Changed on retry" fits better (implying "Tests changed on retry"). Maybe an even better alternative exists.

#### Test-level indicator

The indicator itself is good. There is one minor hiccup though.

Imagine a test with two retries. As long as all retries pass, the test is displayed with the retries indicator and with the total number of retries:

Image

But as soon as we have at least one failed/broken attempt, we lose the information about the number of retries:

Image

But I don't have a good solution in mind. I guess we can left it as is for now. Perhaps, we shouldn't have added these number to begin with.

formaceft-93:
> #### Summary title
> I have two issues with the summary title:
> 
> Image
> 1. The text is quite long.
> 2. It reads as we're counting status transitions across retries rather than tests with at least one such transition.
> 
> I don't have a perfect text in mind. Something like "Changed on retry" fits better (implying "Tests changed on retry"). Maybe an even better alternative exists.
> 
> #### Test-level indicator
> The indicator itself is good. There is one minor hiccup though.
> 
> Imagine a test with two retries. As long as all retries pass, the test is displayed with the retries indicator and with the total number of retries:
> 
> Image
> But as soon as we have at least one failed/broken attempt, we lose the information about the number of retries:
> 
> Image
> But I don't have a good solution in mind. I guess we can left it as is for now. Perhaps, we shouldn't have added these number to begin with.

Thanks, agree on both points.

I shortened the summary label to “Changed on retry”, and updated the tree indicator so it no longer replaces the retry count. Now tests with changed retry status show both pieces of info: the usual retry count and the status-change marker.

PR #979, opened 2026-09-15
add dump field to config?

### LINKED ISSUE #970: Config-level `dump` option is rejected as unsupported
### What happened?

The multistage-build [documentation](https://allurereport.org/docs/multistage-builds/#using-the-config-file) says a dump archive can be produced by
  setting `dump` in the Allure configuration:

  ```js
  export default {
    dump: "./snapshots/stage_1",
  };
```

  However, Allure 3.17.0 rejects this configuration before processing the
  results.

### How can we reproduce it?

1. Create an allure-results directory containing a valid result.
  2. Create allurerc.mjs:

  export default {
    dump: "./snapshots/stage_1",
  };

  3. Run:

  npx allure@3.17.0 generate ./allure-results --config ./allurerc.mjs

### What did you expect?
Allure should process the results, create ./snapshots/stage_1.zip, and skip report generation as documented. Instead I got:

Internal Error: The provided Allure config contains unsupported fields: dump
No dump archive is created.

### Environment

- Allure: 3.17.0
  - Node.js: 22.23.2
  - OS: Linux

### Additional context

 dump is declared in the public Config interface:

  https://github.com/allure-framework/allure3/blob/v3.17.0/packages/plugin-api/src/config.ts#L24-L28

  However:

  1. dump is absent from supportedFields in validateConfig():

  https://github.com/allure-framework/allure3/blob/v3.17.0/packages/core/src/config.ts#L160-L183

  2. resolveConfig() does not propagate config.dump into the returned
     FullConfig.

  3. AllureReport already supports the resolved dump value and writes the
     archive instead of generating a report.

  Git history suggests this omission dates from the original multistage
  implementation in #309. The option was originally named stage and was
  renamed to dump in #504, but neither version was added to the config
  validation/resolution path.

  The CLI allure run --dump path works because that command applies its CLI
  option directly when constructing AllureReport, bypassing the missing
  config propagation.

### PR DESCRIPTION
**Pull request by formaceft-93**
fixed https://github.com/allure-framework/allure3/issues/970

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### CODE REVIEW

File: packages/core/src/config.ts [hunk 1]
```diff
@@ -382,6 +384,7 @@ export const resolveConfig = async (config: Config, override: ConfigOverride = {
     historyLimit,
     historyPath: historyPath ? resolve(historyPath) : undefined,
     historyBaseUrl,
+    dump,
```
epszaw, line 387: Run сommand overwrites `config.dump` with undefined when `--dump` is omitted, so config-only dump still generates report. Use `dump: this.dump ?? config.dump`.

epszaw (changes requested):

epszaw (approved):

PR #984, opened 2026-09-16
attachment copy button

formaceft-93:
## Summary

Adds a copy-to-clipboard control for text-like attachments in the Awesome report.

The copy button is rendered by the shared `AttachmentCode` component, so it applies consistently to JSON, text, URI list, HTML/Markdown source, CSS/code, and ANSI/code attachment views. The control is positioned as a bottom-right overlay and slides in from below on hover/focus, keeping the attachment content uncluttered by default.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/web-components/src/components/Attachment/AttachmentCode.tsx
```diff
@@ -106,35 +129,61 @@ export const AttachmentCode = (props: AttachmentProps & { highlight?: boolean })
...
+  const copyAction = canCopy && (
+    <div className={styles["attachment-code-actions"]}>
+      <TooltipWrapper tooltipText={copyLabel} tooltipTextAfterClick={copiedLabel}>
+        <IconButton
+          style="raised"
+          size="s"
+          iconSize="s"
+          icon={allureIcons.lineGeneralCopy3}
+          onClick={(e: MouseEvent) => {
+            e.stopPropagation();
```
todti: `TooltipWrapper` registers its `onClick` on its own outer wrapper `<div>` and relies on the click bubbling up from the child (`Tooltip/index.tsx#L191`). `stopPropagation()` here prevents that, so `setCurrentText(tooltipTextAfterClick)` never runs and the "Successfully copied" state never appears, which makes the `controls.clipboardSuccess` key that's plumbed through `I18nProp` dead code.

Verified on a local build: the copy itself works (26 137 chars from a 400-line log land in the clipboard), but the tooltip still reads "Copy to clipboard" after the click. The copy button in the error block right above uses the same `TooltipWrapper` + `IconButton` pair without `stopPropagation`, and it does switch correctly.

Moving the guard one level up still keeps the click off the attachment row, but lets the tooltip see the event:

```tsx
<div className={styles["attachment-code-actions"]} onClick={(e: MouseEvent) => e.stopPropagation()}>
  <TooltipWrapper tooltipText={copyLabel} tooltipTextAfterClick={copiedLabel}>
    <IconButton ... onClick={() => copyToClipboard(rawText)} />
  </TooltipWrapper>
</div>
```

Rebuilt with this change, and the tooltip switches as intended.

packages/web-components/src/components/Attachment/AttachmentCode.tsx
```diff
@@ -79,8 +83,27 @@ const shouldShowLineNumbers = (prismLang: string, rawText: string): boolean => {
   return rawText.split("\n").length >= 5;
 };
 
+const copyToClipboard = async (text: string) => {
```
todti: This is the fourth copy of the same helper:

- `packages/web-components/src/components/Attachment/HttpAttachment.tsx#L319`: same directory, identical apart from the missing `focus()`
- `packages/web-awesome/src/utils/copyToClipboard.ts`: identical
- `packages/web-classic/src/utils/copyToClipboard.ts`: identical

`@allurereport/web-commons` is already a dependency of both `web-components` and `web-awesome` (`Attachment.tsx` imports from it), so it looks like the natural home: one implementation, four call sites, three deletions.

Separately, `catch {}` swallows every failure while the tooltip still reports success. `controls.clipboardError` already exists in the locales and is currently unused.

packages/web-components/src/components/Attachment/styles.scss
```diff
@@ -27,6 +27,32 @@
...
+  &:focus-within {
+    .attachment-code-actions {
+      opacity: 1;
+      pointer-events: auto;
+      transform: translateY(0);
+    }
+  }
+}
+
+.attachment-code-actions {
```
todti, line 34: `position: absolute; bottom: 8px` pins the button to the bottom of the code block, and inline attachments have no height cap. Measured on a local build at 1440x900:

- `server.log`, 400 lines (25 KiB): the block is 7232 px tall
- `rows.json`, 800 lines (66 KiB): 18 104 px

The hover reveal itself works anywhere in the block, but the button is painted far below the viewport, so copying a long log means scrolling to the very end first.

`sticky` works here. The scrolling ancestor is nine levels up and every element in between is `overflow: visible`:

```scss
.attachment-code-actions {
  position: sticky;
  top: 8px;
  float: right;
  margin: 8px 8px 0 0;
}
```

with `copyAction` rendered as the first child of `.attachment-code` (`sticky` needs the element in flow, `float` keeps it from pushing the code down). Tried it on the same 400-line log: scrolled 3500 px in, the button stays 8 px below the container's top edge.

Top-right would also line up with the copy button in the error block just above, which is top-right and always visible. Right now the two sit in opposite corners of the same screen.

packages/web-components/src/components/Attachment/Attachment.tsx
```diff
@@ -111,11 +111,12 @@ export const Attachment = (props: AttachmentTestStepResultProps) => {
   const CurrentPreviewComponent = previewComponentsByAttachmentType[componentType];
   const CurrentComponent = componentsByAttachmentType[componentType];
 
-  // @ts-expect-error TODO: add all translations for attachment types
-  const i18nProp = i18n?.[componentType === "image-diff" ? "imageDiff" : componentType];
-
   if (DUAL_VIEW_ATTACHMENT_TYPES.has(componentType) && CurrentPreviewComponent && CurrentComponent) {
     const isCodeComponent = CurrentComponent === AttachmentCode;
+    const i18nProp = isCodeComponent
```
todti: The same ternary is now repeated three times (here, L142, L152), each carrying its own `@ts-expect-error`. Computing it once before the branches covers all three:

```ts
// @ts-expect-error TODO: add all translations for attachment types
const typeI18n = i18n?.[componentType === "image-diff" ? "imageDiff" : componentType];
const i18nFor = (Component: unknown) => (Component === AttachmentCode ? i18n?.controls : typeI18n);
```

It also drops a side effect of the current shape: in the dual-view branch `isCodeComponent` is true for `html`/`markdown`, so `i18n?.controls` is passed to `HtmlPreview`/`MarkdownPreview` too. Harmless today since neither reads `i18n`, but it's a trap for whoever wires them up.

todti (approved):

epszaw (approved):

PR #991, opened 2026-09-17
Keep rerun reports aligned with the final attempt

epszaw:
### Context

When `allure run --rerun` retries a failed test and a later attempt passes, the generated report now reflects the final process attempt. Process errors and stdout/stderr attachments from earlier failed attempts no longer remain in an otherwise successful report.

If the final attempt fails, its diagnostics remain available. Global errors and attachments produced independently by test integrations are preserved across reruns.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/plugin-agent/src/plugin.ts
```diff
@@ -4010,23 +4010,15 @@ export class AgentPlugin implements Plugin {
...
-          await appendRuntimeEvent(runtime, "run_error", {
-            message: error.message ?? "Captured global error",
-          });
-        });
-      }),
-    );
+      });
+    };
+
+    runtime.unsubscribers.push(realtime.onGlobalError(onGlobalError));
```
formaceft-93, line 4021: `executeAllureRun` now publishes test-process failures with `sendProcessGlobalError`, but the realtime agent still subscribes only to `onGlobalError`. That means live agent output will miss these runner/process errors in `manifest/test-events.jsonl` until finalization. Should we also register `realtime.onProcessGlobalError(onGlobalError)` here, and add a test that emits a process global error through the realtime mock?

formaceft-93 (approved):

packages/plugin-agent/src/plugin.ts
epszaw: I think it's enough to handle only global errors, not process errors in realtime mode, due to the fact that in realtime mode we don't expect the process to end but work continuously.

